PR #20, opened 2013-09-01
Switch to Autotools for build and add a debian package creator.

EddyBeaupre:
Convert the build system to autoconf/automake, not that it was really needed since the whole build process is quite simple, but it make things easier to create packages from it. Add a dpkg-deb target to create a debian package and also add a simple man page for quick reference.


n1tehawk:
This PR is obsolete; the commits were merged as part of the [autotools branch](https://github.com/linux-sunxi/sunxi-tools/tree/autotools).

PR #31, opened 2014-11-27
Reorganize folder structure, improve build procedure

semlanik:
Hi,
I improved build procedure, without using any autotools etc.
In README there are build insturctions


ssvb:
This old pull request did not seem to have moved anywhere and now even can't be applied to the current git. Should we just close it?

PR #32, opened 2015-06-30
Add readmem to read physical memory

rgwan:
Some device can't use FEL to get script.bin out.
So I write this simple tool to directly get the script.bin out.
Just to ./readmem 0x43000000 0x1000 > script.bin and you will got it.
It only works at SUNXI specific kernel 3.4 or lichee kernel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ CFLAGS += -std=c99 -D_POSIX_C_SOURCE=200112L
 CFLAGS += -Iinclude/
 
 TOOLS = fexc bin2fex fex2bin bootinfo fel pio
-TOOLS += nand-part
+TOOLS += nand-part readmem
 
 MISC_TOOLS = phoenix_info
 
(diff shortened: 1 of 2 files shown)


ssvb:
This functionality is already provided by https://github.com/linux-sunxi/sunxi-tools/blob/master/script_extractor.c
It needs to be better documented though.

oliv3r:
Yeah, it was a quick and dirty hack :) So improvements of all sorts are 
more then welcome :)

On 17-08-15 18:52, Siarhei Siamashka wrote:
>
> This functionality is already provided by 
> https://github.com/linux-sunxi/sunxi-tools/blob/master/script_extractor.c
> It needs to be better documented though.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/linux-sunxi/sunxi-tools/pull/32#issuecomment-131887699>.
>

-- 
Met vriendelijke groeten, Kind regards, 与亲切的问候

Olliver Schinagl
Research & Development
Ultimaker B.V.

ssvb:
@oliv3r Sure. Fore example, adding a description of this tool to the readme file :) So that @rgwan could have a chance to notice and use it instead of reimplementing something of his own.

PR #33, opened 2015-10-21
added option --progress which shows a simple progress bar

kaplan2539:
Hey,

I added the option '--progress' or '-p' to show some progress bar on the screen while uploading large files via fel write. Hope this is useful.

Cheers
Alex

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -64,6 +64,7 @@ static int AW_USB_FEL_BULK_EP_OUT;
 static int AW_USB_FEL_BULK_EP_IN;
 static int timeout = 60000;
 static int verbose = 0; /* Makes the 'fel' tool more talkative if non-zero */
+static int progress = 0; /* Makes the 'fel' tool print out '.' for each byte transferred */
 static uint32_t uboot_entry = 0; /* entry point (address) of U-Boot */
 static uint32_t uboot_size  = 0; /* size of U-Boot binary */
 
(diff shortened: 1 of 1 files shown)


ssvb:
Hi,

Usually patches are sent to the linux-sunxi mailing list with the "[PATCH sunxi-tools M/N]" prefix in the subject line. This can be done with the help of "git format-patch" and "git send-email" commands.

Regarding your patch, there are some things that might need to be fixed:
1. Inconsistent coding style. The rest of the code is using spaces to separate brackets and operators, for example we use "if (...)" instead of "if(...)".
2. The new '-p' command line option is not documented in the "help" message.
3. The '-v' and '-p' options may not be processed correctly if they are used together, depending on their order.
4. It would be probably cleaner to add a progress callback function argument to the "usb_bulk_send()" function instead of doing all the progress bar animation there directly. Because this is a low level helper function and it has a wider use than just uploading large files when handling one or more "write" command line options.

kaplan2539:
1. OK - I'll improve that
2. OK - Will be added
3. I don't understand - according to my tests they should be processed correctly independent of their order
4. I'll do.

ssvb:
> I don't understand - according to my tests they should be processed correctly independent
> of their order

For example, what happens if you order the command line arguments as 'fel -p -v ver' ?

kaplan2539:
got it - whould it make sense to use getopt?

kaplan2539:
followed your suggestions and tested a bit.
This time I sent the patch to the sunxi-linux mailing list.

PR #36, opened 2015-12-29
Complete rework of sunxi-nand and new autotool conversion.

EddyBeaupre:
What started as a simple change to add some sanity to the command line parser of sunxi-nand, ended up being an almost complete rewrite of it. The only original code left are some headers and crc calculator (with some modifications).

Sunxi-nand is now capable of adding/deleting/editing partitions, it can also create a brand new MBR. It work on device or image file (without having to attach them to a loop device). It can also do silly things like writing a SUN4I MBR on SUN7I device if that's really what you want to do.

I know i did an autotool conversion long ago, but it kinda felt behind master quite alot. So i redid that work too. It can cross-compile almost everything except sunxi-fel which would need to install the arm version of libusb in a foreign system architecture. Not that is is terribly hard to do but i can see the trail of discutions about how to do it comming to the mailing list and the tons of RTFM comment that would follow.

To follow what was started earlier, everything has been renamed sunxi-, so it will be easier to have the tools include in officials repo.

One little change to sunxi-fel, it now only check if parts of its name is fex2bin or bin2fex to decide its mode of operation instead of an exact match. Help is also a little more precise.

The /bin directory is gone since everything there can be made by the Makefile.

Make also have an help target, and a sunxi-bin target to build the bare-metal things.

mksunxiboot is also include (rename to sunxi-mkboot), so the targets sunxi-jtag-loop.sunxi and sunxi-fel-sdboot.sunxi can now build directly.


ssvb:
Thanks for working on improving sunxi-tools. We don't use github pull requests though. The patches should be sent to the linux-sunxi mailing list for review (via "git format-patch" and "git send-email") with the subject lines edited to include the "sunxi-tools" tag. For example, they could look like "[PATCH sunxi-tools 1/3] nand-part: Fix XYZ problem".

Regarding the patches:
1. We don't want to have "cleanup" patches. Just earlier patches in the series should try no to introduce problems in the first place.
2. Please don't pile up multiple unrelated changes. If your primary goal is to improve the sunxi-nand-part program, then please do it without the autotools stuff.
3. It also makes sense to improve the documentation at http://linux-sunxi.org/Sunxi-tools#sunxi-nand-part (frankly speaking, I have never used sunxi-nand-part myself and would appreciate a bit more information about it).

ssvb:
> It also makes sense to improve the documentation at http://linux-sunxi.org/Sunxi-tools#sunxi-nand-part (frankly speaking, I have never used sunxi-nand-part myself and would appreciate a bit more information about it).

There are some other places in the linux-sunxi wiki, which mention the nand-part tool too:
http://linux-sunxi.org/NAND#nand-part
http://linux-sunxi.org/Installing_to_NAND#nand-part

Basically, what I mean is that all these instructions in the wiki should be valid and accurately describe the nand-part tool. If you are changing the behaviour of the nand-part tool (we should generally avoid doing this unnecessarily) or introducing new features, then all of this should be documented too. Also until the new features are accepted in the official sunxi-tools git repository, the wiki pages should clearly state that these features are still only available in your repository, with a link to this repository.

EddyBeaupre:
I agree and I also plan to make proper man pages for all the tools.


Envoyé depuis mon appareil Samsung

-------- Message d'origine --------
De : Siarhei Siamashka <notifications@github.com> 
Date : 2015-12-29  14:39  (GMT-05:00) 
À : linux-sunxi/sunxi-tools <sunxi-tools@noreply.github.com> 
Cc : Eddy Beaupré <eddy@beaupre.biz> 
Objet : Re: [sunxi-tools] Complete rework of sunxi-nand and new autotool
  conversion. (#36) 


It also makes sense to improve the documentation at http://linux-sunxi.org/Sunxi-tools#sunxi-nand-part (frankly speaking, I have never used sunxi-nand-part myself and would appreciate a bit more information about it).


There are some other places in the linux-sunxi wiki, which mention the nand-part tool too:

http://linux-sunxi.org/NAND#nand-part

http://linux-sunxi.org/Installing_to_NAND#nand-part

Basically, what I mean is that all these instructions in the wiki should be valid and accurately describe the nand-part tool. If you are changing the behaviour of the nand-part tool (we should generally avoid doing this unnecessarily) or introducing new features, then all of this should be documented too. Also until the new features are accepted in the official sunxi-tools git repository, the wiki pages should clearly state that these features are still only available in your repository, with a link to this repository.

EddyBeaupre:
I have think about all this and I am no longer sure merging this with the official tools is the best way to update theses tools. Another goal I have in the near future is to merge sunxi-fex and maybe other tools with sunxi-nand and have a single binary that can do all theses things.

Keeping it a completely separated fork and renaming the final binary to sunxi-tools would probably be less disturbing for the community than having to update tons of documentations.

I think it

EddyBeaupre:
I think it's the best way to go.

Issue #38, opened 2016-04-04
Support a "sid" command for sunxi-fel


n1tehawk:
Many Allwinner SoCs feature a series of efuses that can serve as a unique chip identifier. On most devices these can be read via register access, so in theory our FEL utility could easily retrieve them. We're particularly interested in the 128-bit "SID" key - see http://linux-sunxi.org/SID.

There are SoC-specific differences in how the SID can be accessed (e.g. individual register base addresses). The idea is to add a new "sid" command to our `sunxi-fel` utility, have it handle these differences internally and print out the resulting 128 bits, to provide a simple and uniform way for users to retrieve the SID.

Note: A [shortcoming of the BROM FEL code](https://github.com/ssvb/sunxi-tools/commit/b4116008efe28016ca85de83a7e1d453a1d7e745) won't allow simple retrieval of the hardware registers. Therefore currently `aw_fel_read()` can't retrieve the SID successfully. We first need to implement the proposed `aw_fel_readl()`/`aw_fel_writel()` functions before we can move forward on this issue.

See also: discussion of #37.

n1tehawk:
@ssvb I'd really like to move forward on this one. To do so, I suggest cherry-picking the `scratch_addr` and `readl`/`writel` changes from your A64 branch, and then add support for the `sunxi-fel sid` command on top of it.

I have collected a suitable series of commits in [this branch](https://github.com/n1tehawk/sunxi-tools/commits/sid-command), and would like your comment on it. If you consider that a reasonable approach, then I'd submit the corresponding patches on the mailing list (for peer review / ACK).

@ThomasKaiser The above branch could use some testing on your zoo of sunxi devices... :grinning:

Regards, NiteHawk

n1tehawk:
Closed by merging #41

PR #41, opened 2016-05-03
fel: Add new readl/writel functionality and a "sid" command

n1tehawk:
See https://groups.google.com/forum/#!topic/linux-sunxi/5tvYnr9TBqg and #38

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -662,6 +662,80 @@ void aw_write_arm_cp_reg(libusb_device_handle *usb, soc_sram_info *sram_info,
 	aw_fel_execute(usb, sram_info->scratch_addr);
 }
 
+/* multiple "readl" from sequential addresses to a destination buffer */
+void aw_fel_readl_n(libusb_device_handle *usb, uint32_t addr,
+		    uint32_t *dst, size_t count)
+{
+	soc_sram_info *sram_info = aw_fel_get_sram_info(usb);
+	uint32_t val;
+	uint32_t arm_code[] = {
+		htole32(0xe59f0010), /* ldr        r0, [pc, #16]            */
+		htole32(0xe5901000), /* ldr        r1, [r0]                 */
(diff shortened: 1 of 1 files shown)


fel.c
```diff
@@ -662,6 +662,80 @@ void aw_write_arm_cp_reg(libusb_device_handle *usb, soc_sram_info *sram_info,
...
+	aw_fel_write(usb, arm_code, sram_info->scratch_addr, sizeof(arm_code));
+	aw_fel_execute(usb, sram_info->scratch_addr); /* stores first value */
+	while (--count > 0) {
+		/*
+		 * Subsequent transfers only need to set up the next value
+		 * to store (since the scratch code auto-increments addr).
+		 */
+		aw_fel_write(usb, src++, sram_info->scratch_addr + 28, sizeof(uint32_t));
+		aw_fel_execute(usb, sram_info->scratch_addr);
+	}
```
ssvb: We are likely not getting any performance improvements here by doing this auto-increment trick because the number of round-trips over USB is still the same (we need to write something to the SRAM and then execute it).

n1tehawk: I agree that the difference shouldn't be big. But even with the trivial "sid" retrieval I'm repeatedly seeing a difference of roughly 30ms between a single `aw_fel_readl_n(usb, soc_info->sid_addr, key, 4)` and four sequential `aw_fel_readl()`.

As "going that extra mile" required only moderate effort in this case, I thought I'd opt for it. Maybe the multiple transfers will be useful for some future task.

Regards, NiteHawk

n1tehawk: If the need arises, we could also later extend `aw_fel_writel_n()` to make use of an intermediate SRAM buffer (e.g. the `thunk_addr`?) for bigger, 'chunked' transfers - reducing on the required number of USB exchanges.

PR #42, opened 2016-05-06
Make fel-sdboot.sunxi usable on A64 and A80

ssvb:
Do runtime SoC type detection in ``fel-sdboot.sunxi`` to add A64/A80 support. Untested on A80, but maybe @jemk can help?


fel-sdboot.c
```diff
@@ -36,5 +36,18 @@ dd if=fel-boot.sunxi of=/dev/sdX bs=1024 seek=8
 
 void _start(void)
 {
-	((void (*)(void))0xffff0020)();
+	volatile unsigned int *sram_ver_reg = (void *)0x01C00024;
```
n1tehawk: Can you rename this to SYS_CTL_VER_REG? I find the "sram" slightly misleading.

jemk:
It doesn't work on A80, because there the version register is at 0x00800024 instead of 0x01c00024.
Wouldn't it be possible and more future-proof to check the ARM SCTLR.V bit instead?

ssvb:
@jemk This is a bad news, at least all the other SoC variants used to be consistent. Yes, thanks for the good idea, we can use the SCTLR.V bit.

n1tehawk:
@ssvb Siarhei, I've cherry-picked (and commited) your first commit on top of bbdae674400534ef433a8e14d8e603bdc6e84daa - so you can/should drop it when updating this pull request.

Regards, NiteHawk

ssvb:
Thanks. I have updated the pull request with the new implementation which checks SCTLR.V, hopefully it works on A80 now. I also have encountered some really odd problem and had to come up with a workaround for it.

ssvb:
An example of a non-working fel-sdboot.sunxi - https://gist.github.com/ssvb/5f68caade9c76fd671f68135a6d96643

fel-sdboot.c
```diff
@@ -36,5 +36,24 @@ dd if=fel-boot.sunxi of=/dev/sdX bs=1024 seek=8
...
+	 * FEL mode fails to activate in an unpredictable way without
+	 * this NOP padding. Minor changes in the code, such as checking
+	 * the PC register (PC >= 0x10000) instead of SCTLR.V or doing
+	 * jump instead of call to the FEL handler in the BROM sometimes
+	 * break on A64 and sometimes break on A10/A13/A20. Trying to
+	 * add DSB & ISB instructions and/or invalidating caches and
+	 * BTB do not seem to make any difference. Only adding a bunch
+	 * of NOP instructions in the beginning helps.
+	 */
+	asm volatile(".rept 32 \n nop \n .endr");
```
n1tehawk: nitpick: Personally, I don't like the `\n` characters here; maybe that could be changed to
`asm volatile(".rept 32; nop; .endr");` or `asm volatile(".rept 32 ; nop ; .endr");` ?

fel-sdboot.c
```diff
@@ -36,5 +36,24 @@ dd if=fel-boot.sunxi of=/dev/sdX bs=1024 seek=8
...
+
+	asm volatile("mrc p15, 0, %0, c1, c0, 0" : "=r" (sctlr));
+
+	if (sctlr & (1 << 13)) /* SCTLR.V */
+		((void (*)(void))0xffff0020)();
+	else
+		((void (*)(void))0x00000020)();
```
n1tehawk: What do you think of introducing a macro for more readability? Something along the lines of:
```c
/* treat a given address as function pointer and call (= execute) it */
#define JUMP(addr)     ((void (*)(void))(addr))()
```

fel-sdboot.c
ssvb: "info as" says
<pre>3.5 Statements
==============

A "statement" ends at a newline character (`\n') or a "line separator
character".  The line separator character is target specific and
described in the _Syntax_ section of each target's documentation.  Not
all targets support a line separator character.  The newline or line
separator character is considered to be part of the preceding
statement.  Newlines and separators within character constants are an
exception: they do not end statements.</pre>
And in the ARM specific section:
<pre>The `;' character can be used instead of a newline to separate statements.</pre>
However for some architectures ';' means a comment until the end of line.

fel-sdboot.c
ssvb: I'm not sure. The existing code is IMHO already quite readable for anyone familiar with the C language. And we only have two lines with this code.

jemk:
Works on A80 too.

ssvb:
BTW, this was the root cause of the problems - https://patchwork.ozlabs.org/patch/622171/

PR #43, opened 2016-05-12
fel: Add fel spl command support for Allwinner A64

ssvb:
The SCTLR bits are somewhat different because the V bit is set
to 0 on A64 (Low exception vectors, base address 0x00000000) and
the UNK bit (Reads of this bit return an UNKNOWN value) is also not
the same as on the other SoCs. So the SCTLR check can be relaxed.

Changes in v2:
 - Because the SRAM A and SRAM C reside back-to-back in the address
   space, it is possible to use 40 KiB of SRAM by the SPL for its
   code+data+stack. So the FEL backup storage is moved from 0x18000
   to 0x1A000 to support this.

Signed-off-by: Siarhei Siamashka <siarhei.siamashka@gmail.com>

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -258,6 +258,7 @@ void aw_fel_print_version(libusb_device_handle *usb)
 	case 0x1633: soc_name="A31"; break;
 	case 0x1651: soc_name="A20"; break;
 	case 0x1650: soc_name="A23"; break;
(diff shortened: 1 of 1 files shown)


n1tehawk:
FEL tested successfully on Pine64+ by starting Allwinner's boot0 via USB.

Issue #39, opened 2016-04-04
Support a "--sid" option for device selection


n1tehawk:
We have incorporated some [support for multiple USB devices](https://github.com/linux-sunxi/sunxi-tools/commit/472ac4759df557c00248e557beb869f4fe7d75f7) connected to the same host in our `sunxi-fel` utility, allowing to select a specific device via USB bus and device number (`--dev` option).

As proposed in #37, an even more user-friendly way might be to support selection based on some unique device identifier; for example the SID. This could be done by adding a corresponding `--sid <pattern>` option. The details of how to match _pattern_ haven't been elaborated yet. (substring, regex, ...?) 

Right now the `fel.c` code is not prepared well for this enhancement. Historically, we've always created / selected a libusb handle **once** in a preparatory step, and then kept using it for an entire `sunxi-fel` "session". This needs to be rethought if we want to query several devices "in turn" (involving individual FEL communication to retrieve the ID each time), until we find the one that the user actually requested.

See also: #38

n1tehawk:
Note: Implementing ``sunxi-fel sid` should be a higher priority than this. With working SID output (e.g. usable for `grep`), some scripting might serve as a workaround to select the desired device.

This would print the SIDs of all connected FEL devices:
``` shell
#!/bin/sh
for dev in `lsusb | awk '/1f3a:efe8/ { printf "%s:%.3s ", $2, $4 }'`; do
        echo USB device $dev
        ./sunxi-fel --dev=$dev sid
done
```

ssvb:
BTW, the existing ``--dev`` option does not seem to be very useful to me because the usb id keeps changing after re-plugging the device. So it would be very useful to have the ``--sid`` option implemented. The wrapper script works but is a little bit awkward to use, so it's better to have everything nicely supported by the sunxi-fel tool itself.

n1tehawk:
Implemented in 448fa5f74c0db5c359b72ef4770a8bb84309c947.

PR #47, opened 2016-05-13
Make use of FEL library to implement device enumeration and selection by SID

n1tehawk:
This allows code to be possibly reused, and is keeping better focus on the actual FEL functionality in `fel.c`.

---

Changes in [v1](https://github.com/linux-sunxi/sunxi-tools/commits/959de2eddb1f20f5b6ab2cbdf9f087ef6a6ea183): Initial version

Changes in [v2](https://github.com/linux-sunxi/sunxi-tools/commits/02faa6669622b07ef23af514af2499fb0fac182f):
- rebased on current master (9febb55)
- picked up the suggestion to incorporate _soc_sram_info_ (and FEL version data) into the handle

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -963,10 +963,35 @@ static unsigned int file_upload(feldev_handle *dev, size_t count,
 	return i; /* return number of files that were processed */
 }
 
+static void felusb_list_devices(void)
+{
+	size_t devices; /* FEL device count */
+	feldev_list_entry *list, *entry;
+
+	list = list_fel_devices(&devices);
+	for (entry = list; entry->soc_version.soc_id; entry++) {
+		printf("USB device %03d:%03d   Allwinner %-8s",
+			entry->busnum, entry->devnum, entry->soc_name);
+		/* output SID only if non-zero */
+		if (entry->SID[0] | entry->SID[1] | entry->SID[2] | entry->SID[3])
+			printf("%08x:%08x:%08x:%08x",
+			       entry->SID[0], entry->SID[1], entry->SID[2], entry->SID[3]);
+		putchar('\n');
+	}
+	free(list);
+
+	if (verbose && devices == 0)
+		fprintf(stderr, "No Allwinner devices in FEL mode detected.\n");
+
+	feldev_done(NULL);
+	exit(devices > 0 ? EXIT_SUCCESS : EXIT_FAILURE);
+}
+
 int main(int argc, char **argv)
 {
 	bool uboot_autostart = false; /* flag for "uboot" command = U-Boot autostart */
 	bool pflag_active = false; /* -p switch, causing "write" to output progress */
+	bool device_list = false; /* -l switch, prints device list and exits */
 	feldev_handle *handle;
 	int busnum = -1, devnum = -1;
 
@@ -975,6 +1000,7 @@ int main(int argc, char **argv)
 		printf("Usage: %s [options] command arguments... [command...]\n"
 			"	-v, --verbose			Verbose logging\n"
 			"	-p, --progress			\"write\" transfers show a progress bar\n"
+			"	-l, --list			Enumerate all (USB) FEL devices and exit\n"
 			"	-d, --dev bus:devnum		Use specific USB bus and device number\n"
 			"\n"
 			"	spl file			Load and execute U-Boot SPL\n"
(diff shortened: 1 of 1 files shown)


ssvb:
Yes, some change of the code structure would be useful. Regarding the use of a pointer to this new structure as a handle, I wonder if it makes sense to also encapsulate the ``soc_sram_info`` pointer there and get rid of the ``aw_fel_get_sram_info()`` function? The functions, which currently take both ``soc_sram_info`` and ``libusb_device_handle`` pointer arguments, can be changed to just keep one of them (a single ``felusb_handle``?).

Do you have any other patches which make use of this? It is always best to demonstrate the usefulness of code refactoring by adding some new feature, which was a little bit problematic before. Maybe #39 is the one?

n1tehawk:
I have no working code for it yet - but yes, #39 is what i had in mind here. Keeping the low-level stuff (e.g. claiming interface, endpoint detection, etc.) away from the _fel.c_ utility should enable us to safely open and close multiple AW devices in a loop, querying each in turn until we find the desired one.

As for the `soc_sram_info` idea: I'll have to examine that more closely. It might be useful.

ssvb:
Regarding the code splitting, we can also look at the ``fel.c`` source file as the provider of the core FEL protocol handling (the USB stuff), while the other source files may be adding extra functionality. For example, it is natural to introduce [SPI flash support](https://github.com/ssvb/sunxi-tools/commits/20160523-spiflash-wip) via adding new ``fel-spiflash.c`` and ``fel-spiflash.h`` source files.

If we move the USB basics away support from ``fel.c``, then what is remaining there? The command line handling?

> As for the soc_sram_info idea: I'll have to examine that more closely. It might be useful.

It is just more convenient to have a single handle to pass between multiple functions (with all the state information being retrievable from it). With each connected device having its own handle, if we are about to start taking care of multiple devices simultaneously.

n1tehawk:
Hi Siarhei!

> Regarding the code splitting, we can also treat the ``fel.c`` source file as the provider of the core FEL protocol handling (the USB stuff), while the other source files may be adding extra functionality.

Part of what got me started on this _usblib_ series was the notion that "FEL protocol handling" and "the USB stuff" - while still closely related - are actually two different things. It was kind of thinking 'layers' here: The FEL functionality represents a higher level, and is implemented by issuing suitable FEL requests ('protocol') over USB (the 'transport medium'). My goal was to isolate the latter two, and focus ``fel.c`` on the more interesting "FEL" stuff - which in turn relies on / makes use of the library routines.

> If we move the USB basics away from ``fel.c``, then what is remaining there? The command line handling?

[fel.c @ 9e058e9](https://github.com/n1tehawk/sunxi-tools/blob/9e058e911a8a25edfd7403a4bf077f31eccd65d3/fel.c) - Command line handling, plus their actual implementation; basically anything at or above "FEL" level. One prominent example would be the setup and execution of thunk code - e.g. SPL execution.

> It is just more convenient to have a single handle to pass between multiple functions (with all the state information being retrievable from it). And each connected device having its own handle, if we are about to start taking care of multiple devices simultaneously.

Initially I shied away a bit from this idea, as it would mark a first mixup of what I considered "higher" functionality with the low-level routines from `fel_usblib.c`. From a practical point of view however, I do recognize that most of the time we'll probably do a `aw_fel_get_version()` request (and deduct SoC info from the result) sooner or later anyway. So I'm okay with making this part of the "handle" (or state, if you like). But I'd prefer if the different SoC data structures and matchup code from `aw_fel_get_sram_info()` would go into a code unit of their own, instead of integrating them with the _usblib_.

n1tehawk:
I'm now at a point where a more general handle type also provides FEL version and SoC information:
```c
typedef struct _feldev_handle {
	felusb_handle *usb;
	struct aw_fel_version soc_version;
	soc_sram_info *soc_info;
} feldev_handle;
```

I'll have to see when I find some time to progress with an actual "use case", e.g. addressing #39. I would suggest reviewing and merging this series, before moving on with a separate PR - the `--sid` option doesn't fit in here well IMHO...

n1tehawk:
I'm rebasing this series to current master, adding another field for the SoC name to the "device handle".

n1tehawk:
Following the v1.4.2 point release, I have now merged the "FEL library" changes. What remains here is two more commits that demonstrate new use cases by adding a `--list` option and device selection via `--sid`.

PR #49, opened 2016-05-18
fexc fixes

n1tehawk:
I'm currently experimenting with functional tests that do a "`fex2bin` + `bin2fex` cycle" on all FEX files from [sunxi-boards](https://github.com/linux-sunxi/sunxi-boards). In doing that, two minor problems with the current code showed up. This pull request addresses those.

### DIFF
--- a/script_bin.c
+++ b/script_bin.c
@@ -269,9 +269,12 @@ static int decompile_section(void *bin, size_t bin_size,
 			} else if (gpio->port == 0xffff) {
 				; /* port:power */
 			} else if (gpio->port < 1 || gpio->port > GPIO_BANK_MAX) {
-				pr_err("%s: %s.%s: unknown GPIO port bank %c (%u)\n",
-				       filename, section->name, entry->name,
-				       'A'+gpio->port, gpio->port);
+				pr_err("%s: %s.%s: unknown GPIO port bank ",
+				       filename, section->name, entry->name);
+				char c = 'A' + gpio->port - 1;
+				if (c >= 'A' && c <= 'Z')
+					pr_err("%c ", c);
+				pr_err("(%u)\n", gpio->port);
 				goto failure;
 			}
 			v[0] = gpio->mul_sel;
--- a/script_bin.c
+++ b/script_bin.c
@@ -17,6 +17,7 @@
 
 #include "common.h"
 
+#include <ctype.h>
 #include <errno.h>
 #include <stdlib.h>
 #include <stdint.h>
@@ -239,6 +240,13 @@ static int decompile_section(void *bin, size_t bin_size,
 		unsigned type, words;
 		type	= (entry->pattern >> 16) & 0xffff;
 		words	= (entry->pattern >>  0) & 0xffff;
+
+		for (char *p = entry->name; *p; p++)
+			if (!(isalnum(*p) || *p == '_')) {
+				pr_info("Warning: Malformed entry key \"%s\"\n",
+					entry->name);
+				break;
+			}
 
 		switch(type) {
 		case SCRIPT_VALUE_TYPE_SINGLE_WORD: {
--- a/script_bin.c
+++ b/script_bin.c
@@ -318,7 +318,7 @@ int script_decompile_bin(void *bin, size_t bin_size,
 	unsigned int i;
 	struct script_bin_head *head = bin;
 
-	if ((head->version[0] > SCRIPT_BIN_VERSION_LIMIT) ||
+	if (((head->version[0] & 0x3FFF) > SCRIPT_BIN_VERSION_LIMIT) ||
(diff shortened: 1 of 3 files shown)


script_bin.c
```diff
@@ -271,7 +271,7 @@ static int decompile_section(void *bin, size_t bin_size,
 			} else if (gpio->port < 1 || gpio->port > GPIO_BANK_MAX) {
 				pr_err("%s: %s.%s: unknown GPIO port bank %c (%u)\n",
 				       filename, section->name, entry->name,
-				       'A'+gpio->port, gpio->port);
+				       '@'+gpio->port, gpio->port);
```
ssvb: The '@' character looks a little bit odd here. Maybe it's better to have a cleaner error message variant for the cases when the bank number can't be mapped to a valid latin letter?

ssvb:
The "fexc: Don't fail silently on parse error" patch looks good to me. BTW, what are we going to do with the discovered broken FEX files in the sunxi-boards repository?

script_bin.c
n1tehawk: It looks odd, but the message also prints the decimal bank number - so it should be possible for the user to identify the problem even when `gpio->port` <= 0. There are other places where something like `'A' + gpio->port - 1` gets used - but it don't really consider that "cleaner".

n1tehawk:
> BTW, what are we going to do with the discovered broken FEX files in the sunxi-boards repository?

No idea. :smile: I have created corresponding issues, but it remains to be seen whether anybody cares... Who's "in charge" of the _sunxi-boards_ repo? I noticed that there also seem to be several pull requests pending.

One of the reasons I'm aiming at some automated testing of the _.fex_ files is to have a proper functional test of `sunxi-fexc` - and the other is to discover such bad configurations, and prevent  "broken" files from going into the repo again in the future (or a least: discover and fix them early).

n1tehawk:
If that was more of a "technical" question instead of an administrative one: For now I have simply resorted to using a set of patches to the _.fex_ files for my testing - see https://github.com/n1tehawk/sunxi-tools/commit/2876afff9fc674379eb1f7e92c4a430cce76e4fe. With those applied (and the `sunxi-fexc` modifications from this PR), the _.fex &rarr; .bin &rarr; .fex_ cycle can be completed successfully for all _sys_config_ files in the repo.

ssvb:
> Who's "in charge" of the sunxi-boards repo?

Not sure about this, maybe @jwrdegoede knows something?

> One of the reasons I'm aiming at some automated testing of the .fex files is to have a proper functional test of sunxi-fexc - and the other is to discover such bad configurations, and prevent "broken" files from going into the repo again in the future (or a least: discover and fix them early).

Yes, enabling continuous integration for sunxi-boards and doing at least some basic automatic validation for the contributed FEX files would make sense.

jwrdegoede:
Hi,

On 24-05-16 11:32, Siarhei Siamashka wrote:
>     Who's "in charge" of the sunxi-boards repo?
>
> Not sure about this, maybe @jwrdegoede <https://github.com/jwrdegoede> knows something?

No one really is, a bunch of us has access and we all pull in patches posted to
the list every now and then. It is really just a collection of fex files
for easy reference.

Regards,

Hans

n1tehawk:
Hi Hans!

>It is really just a collection of fex files for easy reference.

Some of which are untested / broken, i.e. won't compile with our current `sunxi-fexc`...

@ssvb Siarhei - if I'm not mistaken, you have write access to the sunxi-boards repo, too.

Regards, NiteHawk

n1tehawk:
I have reworked the second commit not to output any 'invalid' bank letters any more (and just use the number instead). From my point of view, this should be good to merge now?

ssvb:
Thanks, looks good now.

n1tehawk:
I'll probably add to this PR to address a quirk detected with decompiling (malformed) `.bin` files.
See https://github.com/linux-sunxi/sunxi-boards/issues/50#issuecomment-221712472

n1tehawk:
In the end I decided to follow a two-fold strategy here: 1. Have `bin2fex` issue a warning when encountering suspicious key names, but otherwise still decompile what's "in" the _.bin_ file; and 2. make `fex2bin` warn on those `':'` lines, but simply ignore them (assuming they were comments originally).

This should solve the problems that https://github.com/linux-sunxi/sunxi-boards/issues/50 related to.

Siarhei, if you're okay with the three new commits above, I'd like to merge this - allowing me to move forward with another branch adding actual _.fex &harr; .bin_ tests to Travis CI.

ssvb:
Thanks, this seems to be reasonable and ``Acked-by: Siarhei Siamashka <siarhei.siamashka@gmail.com>``

n1tehawk:
Merged. Thanks for reviewing!

PR #50, opened 2016-05-24
fel: Implement "rmr" commands to force warm reset into AArch32/AArch64

n1tehawk:
See also: http://irclog.whitequark.org/linux-sunxi/2016-05-23#16539174

The idea is stuff like `sunxi-fel -v spl ${SPL} write 0x44000 ${ATF} rmr64 0x44000`, possibly augmented by suitable (64-bit) U-Boot and kernel binaries (or `u-boot-with-spl.bin` in place of `${SPL}`).

A first test on Pine64 with a debug ATF kindly provided by @apritzel was successful, and demonstrated smooth transition to pure 64-bit code.

---
Changes in [v1](https://github.com/linux-sunxi/sunxi-tools/commits/342457d85ed5e649f7f02960745f49cb5de31bea): Initial version

Changes in [v2](https://github.com/linux-sunxi/sunxi-tools/commits/23f239d0f3a951bd1a4aef6a19c0718a4e308d22):
- Do **not** expose any 32-bit functionality (`reset32` into AArch32) for now; we first need to better understand the requirements and implications.
- Renamed `rmr64` command to `reset64`.

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -465,6 +465,7 @@ typedef struct {
 	bool               needs_l2en;   /* Set the L2EN bit */
 	uint32_t           mmu_tt_addr;  /* MMU translation table address */
 	uint32_t           sid_addr;     /* base address for SID_KEY[0-3] registers */
+	uint32_t           rvbar_reg;    /* MMIO address of RVBARADDR0_L register */
 	sram_swap_buffers *swap_buffers;
 } soc_sram_info;
 
@@ -588,6 +589,7 @@ soc_sram_info soc_sram_info_table[] = {
 		.thunk_addr   = 0x1A200, .thunk_size = 0x200,
 		.swap_buffers = a64_sram_swap_buffers,
 		.sid_addr     = 0x01C14200,
+		.rvbar_reg    = 0x017000A0,
 	},
 	{
 		.soc_id       = 0x1673, /* Allwinner A83T */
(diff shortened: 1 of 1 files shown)


ssvb:
Thanks. This looks like a kind of low level functionality (together with ``readl`` and ``writel`` commands) which is useful for developing code or experimenting with something. If @apritzel approves, then I have no objections and we can push this patch.

Still I guess, the end users would eventually prefer ``"sunxi-fel uboot a-single-binary-with-everything-buindled-in-it"`` for booting Allwinner A64 boards after we have everything sorted out :-)

apritzel:
So I tested rmr64 yesterday and it works like a charm. So thanks for that, it saves me another U-Boot hack.
However I am not so sure about rmr32. The effect should be very similar to the "execute" command, but it didn't work for me. Also I would it more expect to switch from AArch64 into AArch32, for which we would need to inject AArch64 code. I see that we currently cannot test this, since all ARMv8 Allwinner parts so far reset into AArch32.
So I wonder if we:
a) find and fix the bug that prevents rmr32 from working
b) drop rmr32 at all for now

Thoughts?

n1tehawk:
Hi Andre!

I had included the "rmr32" for the sake of completeness - as in theory it's just a single bit (_AA64_ in [RMR_EL3](http://infocenter.arm.com/help/topic/com.arm.doc.ddi0500g/BABDHBEG.html)) that differs from the "rmr64" approach.

But admittedly it seems a bit pointless from 32-bit BROM/FEL mode, and I also didn't find a way to really test it successfully. We might acknowledge this fact, drop "rmr32" and rename "rmr64" to just "rmr"...

@ssvb 
> Still I guess, the end users would eventually prefer ``"sunxi-fel uboot a-single-binary-with-everything-bundled-in-it"`` for booting Allwinner A64 boards after we have everything sorted out :-)

Yes, that's our long-term goal. But even that might still require a 32&rarr;64 bit transition, where the `aw_rmr_request()` might come in handy 'behind the scenes'.

n1tehawk:
P.S.: To be easier to recognize / understand for "end users", we might also do away with the RMR reference and use `reset32` and/or `reset64` for the command name instead?

ssvb:
> Yes, that's our long-term goal. But even that might still require a 32→64 bit transition, where the aw_rmr_request() might come in handy 'behind the scenes'.

Right.

> P.S.: To be easier to recognize / understand for "end users", we might also do away with the RMR reference and use reset32 and/or reset64 for the command name instead?

Yes, ``reset32`` and ``reset64`` seem to be more reasonable names. For the sake of consistency, we can even easily implement the reset functionality on all Allwinner SoCs just using the watchdog unit (in the same way as U-Boot does this).

Is it impossible to override the Reset Vector Base Address for 32-bit reset at all? The A53 manual seems to say that RVBAR *"contains the address that execution starts from after reset when executing in the AArch64 state"*? If yes, then how about just having a ``reset`` command (without any arguments) and a ``reset64`` command (with a single argument for the vector base address).

apritzel:
@ssvb: the RVBAR you are looking at is probably RVBAR_EL3, which is a 64-bit system register. The ARMv8 ARM section G6.2.115 show you the "RVBAR" AArch32 system register, which "contains the IMPLEMENTATION DEFINED address that execution starts from after reset when executing in AArch32 state." Just need to find out if those are actually different registers or if they are aliased.

n1tehawk:
For the time being, I feel inclined to limit this to a `reset64 <addr>` command - which is the part that proved actually useful. (`reset` or `reset32 <addr>` currently have no practical value, and might as well get implemented later.)

apritzel:
@n1tehawk: I agree, just leave the others out for now and we add them later with the right semantics once we can test this.
Though I feel the command name "reset64" is a bit generic and possibly misleading, while "rmr64" is more specific and more correct. But I don't really have a strong opinion on this and maybe biased here.

n1tehawk:
Let's keep it "simple". The `-v` flag (_pr_info_ output) already gives a pretty decent description of what's going on, and those really interested could always peek at the source for `aw_rmr_request()`...

ssvb:
The last 23f239d0f3a951bd1a4aef6a19c0718a4e308d22 revision of the patch is ``Acked-by: Siarhei Siamashka <siarhei.siamashka@gmail.com>``

Later we may probably want to also add a new ``help`` command (something like ``sunxi-fel help reset64``) and/or a linux man page to explain the fine details about the usage of this new ``reset64`` command.

n1tehawk:
Merged. Thanks for reviewing!

Issue #48, opened 2016-05-14
Remove the workaround code from fel-sdboot.c (and possibly uart0-helloworld)


n1tehawk:
The root cause of the problems that surfaced with these ARM binaries is that the U-Boot `mksunxiboot` utility writes a minimal header that is **too small** to avoid corruption of actual code by the BROM bootloader. See https://patchwork.ozlabs.org/patch/622173.

Once this is fixed (in upstream U-Boot), we should probably revert the _NOP_ padding used in `fel-sdboot.c` to circumvent this. The `uart0-helloworld-sdboot.sunxi` under development currently also contains a workaround that won't be needed any more afterwards.

ssvb:
Yes, the workaround needs to be at least changed to branch over the corrupted location. The description at https://github.com/linux-sunxi/sunxi-tools/blob/ce9cf33606492076b81e1157ba9fc54b56379335/fel-sdboot.c#L41-L50 also needs to be updated. It did accurately state the facts, but now we have more information.

As for removing the workaround completely, some people may still try to recompile these binaries and use a buggy ``mksunxiboot`` tool while doing this. It might be safer to keep the workarounds at least for a year or so, maybe even longer.

n1tehawk:
Okay, i'll tag this as "_wontfix_" for the time being - so the issue can be left open until we decide that a sane `mksunxiboot` should have made it everywhere, and the workaround gets finally removed.

n1tehawk:
Note: U-Boot release _v2016.07_ contains the [related commit](http://git.denx.de/?p=u-boot.git;a=commit;h=b19236fd1c1ef289bab9e243ee5b50d658fcac3f) that fixes SPL header size, and thus `mksunxiboot` operation. Images created with this (or a later) version account for the field written by the BROM, i.e. they should no longer suffer from code corruption.

PR #52, opened 2016-05-27
fel-sdboot: Fix header corruption workaround

n1tehawk:
Now that we have a better understanding of what's causing the issue that prevented entering FEL sometimes, we can adjust the workaround code to a proper solution.

This is related to #48.

---
Changes in [v1](https://github.com/linux-sunxi/sunxi-tools/commits/0ae03e9ca2b11ae5d5c3580e50de43156ee88026): Initial version

Changes in [v2](https://github.com/linux-sunxi/sunxi-tools/commits/397f298f2c8dd891ac230e2dbe32d22b730c650d):
* Rewrote the code in assembly, side-stepping the issues brought up in the discussion below
* Updated binary `bin/fel-sdboot.sunxi`

Changes in [v3](https://github.com/linux-sunxi/sunxi-tools/commits/17164d8d9ad4b11ea7e180ca3a8ed1ab042cd950):
* Added "Reviewed-by:" line, and squashed with the updated binary into a single commit


ssvb:
This is not a good fix because there is a prologue code generated by the compiler before the inline assembly block. The #44 code resolves this problem by using a special naked function named "start", which goes to a special ".start" section and is placed in the very beginning by the linker script.

n1tehawk:
Oh. That didn't happen with my toolchain - but i guess it's highly version-dependent. I'll adopt the model from `uart0-helloworld` then...

n1tehawk:
I can reproduce the prologue problem with _armv7a-hardfloat-linux-gnueabi-gcc-4.8.3_. But it seem that
```c
void __attribute__((naked)) _start(void)
{
    /* ... */
}
```
would already solve this?

ssvb:
If you mean to just label the current "main" function as naked, then have a look at https://gcc.gnu.org/onlinedocs/gcc/ARM-Function-Attributes.html#ARM-Function-Attributes

> **naked** This attribute allows the compiler to construct the requisite function declaration, while allowing the body of the function to be assembly code. The specified function will not have prologue/epilogue sequences generated by the compiler. Only basic asm statements can safely be included in naked functions (see Basic Asm). While using extended asm or a mixture of basic asm and C code may appear to work, they cannot be depended upon to work reliably and are not supported.

Mixing C and assembly code is not allowed in naked functions.

n1tehawk:
We'd soon end up with more declarations / framework than actual code for this little bugger. :stuck_out_tongue_closed_eyes: In the end I decided to just rewrite the thing in assembly...

I now get
```asm
sunxi-tools # r2 -a arm -s 0x20 -c 'pd 12' -q fel-sdboot.sunxi
        ,=< 0x00000020      020000ea       b 0x30
        |   0x00000024      00f020e3       nop
        |   0x00000028      00f020e3       nop
        |   0x0000002c      00f020e3       nop
        `-> 0x00000030      100f11ee       mrc p15, 0, r0, c1, c0, 0
            0x00000034      020a10e3       tst r0, 0x2000
            0x00000038      20e0a003       moveq lr, 0x20
            0x0000003c      00e01f15       ldrne lr, [pc, -0]          ; [0x44:4]=0xffff0020  ; 'D'
            0x00000040      1eff2fe1       bx lr
            0x00000044      2000ffff       invalid
            0x00000048      00000000       andeq r0, r0, r0
            0x0000004c      00000000       andeq r0, r0, r0
```
with both gcc **4.8.x** and **4.9.x**; and have tested the resulting file (via SD card) successfully on Banana Pi (A20) and Pine64+ (A64).

ssvb:
Thanks, this makes sense. Not touching the stack at all is another bonus (reduces the possibility of any nasty surprises with future SoCs). Could you please also update the binary in the "bin" directory?

    Reviewed-by: Siarhei Siamashka <siarhei.siamashka@gmail.com>

n1tehawk:
Sure. We might also want to have a peek at `jtag-loop.sunxi`, since it seems to suffer from the same issue. Or do you it exotic enough not to bother?

PR #53, opened 2016-05-30
Extend continous integration by adding tests framework

n1tehawk:
My goal for this series is to implement automated functional tests for `sunxi-fexc`, to ensure that all `.fex` files from the [sunxi-boards](https://github.com/linux-sunxi/sunxi-boards) repository can be compiled to binary format (_fex2bin_) and decompiled back (_bin2fex_) from `.bin` sucessfully.

I'll gradually amend the pull request, to check that the Travis CI status is fine after each commit.

---

[v1](https://github.com/linux-sunxi/sunxi-tools/commits/21461d5a0ebc8f35cd331003177eb7a61b747dd1): Initial version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,9 @@ before_install:
       export TARGET=all;
     fi
 
-# build using the Makefile
+# build (and test) using the Makefile, with a single overall status
 script:
-  - make ${TARGET} && make misc
+  - make ${TARGET} && make misc && make check
 
 # run/simulate a test install
 after_success:
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ PREFIX  ?= /usr/local
 BINDIR  ?= $(PREFIX)/bin
 
 .PHONY: all clean tools target-tools install install-tools install-target-tools
+.PHONY: check
 
 tools: $(TOOLS) $(FEXC_LINKS)
 target-tools: $(TARGET_TOOLS)
@@ -94,6 +95,7 @@ install-misc: $(MISC_TOOLS)
 
 
 clean:
+	make -C tests/ clean
 	@rm -vf $(TOOLS) $(FEXC_LINKS) $(TARGET_TOOLS) $(MISC_TOOLS)
 	@rm -vf version.h *.o *.elf *.sunxi *.bin *.nm *.orig
 
@@ -186,3 +188,6 @@ version.h:
 	@for x in $(TOOLS) $(FEXC_LINKS) $(TARGET_TOOLS) version.h '*.o' '*.swp'; do \
 		echo "$$x"; \
 	done | sort -V > $@
+
+check: $(FEXC_LINKS)
+	make -C tests/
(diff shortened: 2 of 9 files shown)


ssvb:
Nice work. Just a minor nitpick so far. The autotools based projects normally invoke ``"make check"`` to run their test suites. This is the de-facto standard convention, which is well known to Linux distribution maintainers. For more details see https://www.gnu.org/software/automake/manual/html_node/Tests.html

n1tehawk:
Good point. I think for consistency we should also rename the subdirectory from `tests/` to `checks/` then? (Will take care of that a bit later, first I'll have to catch up with latest discussions on the mailing list.)

ssvb:
Ironically, the tests directory is usually called "tests". Don't know, maybe "make tests" implies that the tests are getting compiled, and "make check" implies that they are getting run?

n1tehawk:
Changes in [v2](https://github.com/linux-sunxi/sunxi-tools/commits/cf7fe0dcf882548805fc12ed3af39f83850c3401):
* rebased onto current master (a4557874f03e56fa03fa92c5d14c1f5e6da364a5)
* `make clean` now includes `make -C tests/ clean`
* picked up suggestion to rename Makefile target(s) from `test*` to `check*`

n1tehawk:
Changes in [v3](https://github.com/linux-sunxi/sunxi-tools/commits/42e6b94c9bfeae4018c8e8219e1bd553f0484b97):
* rebased onto current master (89dac0f7eaaedd0d8afa9d5a3c713c7c1ccb9cf6)
* I misread @ssvb's suggestion and implemented `make checks` - this is now fixed to be the intended `make check`. Sorry for the noise...

ssvb:
Looks good in general. Except that the test results depend on the sunxi-boards repository content and may unpredictably break. Maybe for the sunxi-tools tests it is best to use some fixed sunxi-boards snapshot? Or even some reduced set of artificially created FEX files?

And IMHO the sunxi-boards repository should have its own CI integration to prevent bad commits from landing there.

n1tehawk:
I have a feeling that _sunxi-boards_ is a rather "slow moving target"... No one addressed the issues I brought up six weeks ago (and which require some awkward patching here), or cared to respond on the mailing list (my inquiry from June 1st).

This solution may not be perfect, but it's still better than silent breakages going unnoticed. If things come apart, we can still investigate whether the culprit is _sunxi-tools_ or _sunxi-boards_. Just my 0.02€ ...

n1tehawk:
As for CI integration of _sunxi-boards_: Yes, it would be sensible to have that configured too - so that commits and/or pull requests would alert contributors on "bad" submissions. (My idea was that a "build step" of _sunxi-boards_ could check out this repo, and then make use of the .fex testing.)

If we really get to a point where a significantly different set of tests (or too many modfications / exceptions) for _sunxi-boards_ become necessary, I would agree on using a dedicated set of test files, possibly "handcrafted" for specific issues. I simply don't see such a requirement right now.

n1tehawk:
Rebased to v1.4

n1tehawk:
Rebased onto current master. Also added a dedicated build rule that allows for a very simple integration into the `sunxi-boards` repo (see https://github.com/n1tehawk/sunxi-boards/commit/af0252032752c1e35dc56ca48d4763b826d6c518 and https://travis-ci.org/n1tehawk/sunxi-boards/builds/175085349).

PR #54, opened 2016-05-31
Add a tool to generate raw NAND images

bbrezillon:
Generating raw NAND images is particularly useful for boot0 images
creation since the mainline driver is not supporting the funky layout
used by Allwinner's ROM code to load the boot0 binary from NAND.

This tools also allows one to generate raw images for 'normal' partitions
so that they can be flashed before soldering on the NAND on the board
(using a regular NAND programmer).

The tool takes care of generating ECC bytes and randomizing data as
expected by the NAND controller, and re-arranging the ECC/data sections
correctly.

Signed-off-by: Boris Brezillon <boris.brezillon@free-electrons.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ FEXC_LINKS = bin2fex fex2bin
 # Tools which are only useful on the target
 TARGET_TOOLS = sunxi-pio
 
-MISC_TOOLS = phoenix_info
+MISC_TOOLS = phoenix_info sunxi-nand-image-builder
 
 # ARM binaries and images
 # Note: To use this target, set/adjust CROSS_COMPILE and MKSUNXIBOOT if needed
--- a/README.md
+++ b/README.md
@@ -46,6 +46,9 @@ Manipulate PIO register dumps
 ### sunxi-nand-part
 Tool for manipulating Allwinner NAND partition tables
 
+### sunxi-nand-image-builder
+Tool used to create raw NAND images (including boot0 images)
+
 ### jtag-loop.sunxi
 ARM native boot helper to force the SD port into JTAG and then stop,
 to ease debugging of bootloaders.
--- a/nand-image-builder.c
+++ b/nand-image-builder.c
@@ -1,0 +1,1092 @@
+/*
+ * Generic binary BCH encoding/decoding library
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published by
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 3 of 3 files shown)


oliv3r:
This tool used to be called 'spl-image-builder' and was developped with/via C.H.I.P. so it has seen its share of battering there. It's good to see it finding its way into a warm home here now :)

ssvb:
@oliv3r I understand that you are very eager to help and only have good intentions, but patches are normally reviewed before getting pushed to the repository.

Your last commit 7ae3b2cfcb9070fb97f4aaa009a63881ad653d67 ``("Very quick n dirty hack to get a script.bin from /dev/mem")``, which you also just pushed to the repository without bothering to submit it for review, was very broken and caused troubles [for everyone](http://irclog.whitequark.org/linux-sunxi/2015-09-20#14063402) [trying to](http://irclog.whitequark.org/linux-sunxi/2015-11-22#14684979) [use it](http://irclog.whitequark.org/linux-sunxi/2016-02-22#15566956) until finally fixed with 55eec70ceafc4b8b25b4ddcd613c9ca10e41dcf7 fix. Moreover, it never added any kind of readme or usage documentation.

Can I kindly ask you to abstain from rash actions with the sunxi-tools repository in the future? Thanks.

ssvb:
@bbrezillon Do I understand it correctly that significant part of this tool is just a slightly adapted copy of the [bch.c](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/lib/bch.c) file from the Linux kernel? The question is, who is the author of the remaining part of this tool? It seems to be missing any other copyright notices.

Regarding the usage of the tool, it would be nice to have a little bit better documentation. There are tons of configuration parameters. Maybe one or two real life examples could be provided (valid command line options for some of the Allwinner based devices or NAND chip types)?

Also errors reporting is not exactly great. If I just run this tool with some gibberish in the command line, then it simply displays a generic help message which is not very helpful.

bbrezillon:
@ssvb 

Copiright question: Yes, that's why I kept the copiright.

Tool doc/examples question: Sure. Where do you want me to put this documentation (and examples)?
Directly in the help context or do you have a documentation
file/directory?

Error reporting comment: I'll add more error messages, explaining exactly what failed.

BTW, maybe that's because I don't know how to do it in guthub, but I prefer plain email reviews, where you can put your answers just after the questions your were asked ;-).

Thanks for the review.

Boris

oliv3r:
@ssvb back then the repo wasn't even that active, it was just some scribbelings to set aside for archiving. I'm very happy this repository see's high activity right now! But you are right and should have checked up on the activity on this repo before merging Boris's work, as we now more activily seem to be using this repository!

PR #59, opened 2016-10-01
Fix building on FreeBSD

evadot:
Hello,

This patch fix building on FreeBSD.


n1tehawk:
This has resurfaced with Dragonfly BSD (see #83). @evadot could you check if the above patch with an alternative solution would work for you, too?

Regards, NiteHawk

PR #60, opened 2016-10-06
fel: Add the ability to pass uEnv-style data via FEL

n1tehawk:
The patch represents the sunxi-tools side of [this U-Boot change](http://git.denx.de/?p=u-boot.git;a=commitdiff;h=320e0570e67efbd093d7750655a758c66e9d5528), and allows our FEL utility to pass environment data in a _uEnv.txt_-compatible format to a current U-Boot version (v2016.09 or later).
___
For example, use your text editor to save a _my.env_ file with
```
#=uEnv
myvar=world
bootcmd=echo "Hello $myvar."
```
and then test it with
`./sunxi-fel uboot u-boot-sunxi-with-spl.bin write 0x43100000 my.env`

You should see U-Boot's autoboot print the corresponding message and drop you to the prompt, proving that you have successfully overwritten the default "bootcmd".

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -1361,7 +1361,10 @@ void pass_fel_information(libusb_device_handle *usb,
 		pr_info("Passing boot info via sunxi SPL: "
 			"script address = 0x%08X, uEnv length = %u\n",
 			script_address, uEnv_length);
-		uint32_t transfer[] = {script_address, uEnv_length};
+		uint32_t transfer[] = {
+			htole32(script_address),
+			htole32(uEnv_length)
+		};
 		aw_fel_write(usb, transfer,
 			sram_info->spl_addr + 0x18, sizeof(transfer));
 	}
--- a/fel.c
+++ b/fel.c
@@ -1351,16 +1351,19 @@ bool have_sunxi_spl(libusb_device_handle *usb, uint32_t spl_addr)
  * (see "boot_file_head" in ${U-BOOT}/arch/arm/include/asm/arch-sunxi/spl.h),
  * providing the boot script address (DRAM location of boot.scr).
  */
-void pass_fel_information(libusb_device_handle *usb, uint32_t script_address)
+void pass_fel_information(libusb_device_handle *usb,
+			  uint32_t script_address, uint32_t uEnv_length)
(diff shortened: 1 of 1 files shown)


fel.c
```diff
@@ -1351,16 +1351,19 @@ bool have_sunxi_spl(libusb_device_handle *usb, uint32_t spl_addr)
...
 	if (have_sunxi_spl(usb, sram_info->spl_addr)) {
-		pr_info("Passing boot info via sunxi SPL: script address = 0x%08X\n",
-			script_address);
-		aw_fel_write(usb, &script_address,
-			sram_info->spl_addr + 0x18, sizeof(script_address));
+		pr_info("Passing boot info via sunxi SPL: "
+			"script address = 0x%08X, uEnv length = %u\n",
+			script_address, uEnv_length);
+		uint32_t transfer[] = {script_address, uEnv_length};
+		aw_fel_write(usb, transfer,
```
ssvb: This is not big endian friendly.

n1tehawk:
>There is a minor incompatibility with big endian systems. We need to convert all 32-bit words from native byte order to little endian when storing them into structures.

Good catch. We've obviously already missed that before (when just passing the `script_address` alone).

>Also maybe it's a good idea to add the information about the special "#=uEnv" prefix into readme (or into linux-sunxi wiki). Otherwise the users will probably never figure out how it works.

Agreed. For `sunxi-fel` usage the README is pointing to the wiki article on [USBBoot](http://linux-sunxi.org/FEL/USBBoot), so that's probably the place where it should go.

>The SPL header version check is missing. But we probably forgot to bump the header version on the U-Boot side, so nothing can be done about this right now.

The header size did not change in this case, and we're using "safe" defaults on values and behavior - so this isn't a breaking change that would require a header version bump.

The default for `uEnv_length` is 0 (_mksunxiboot_ is expected to clear the header), and won't be touched by older `sunxi-fel` versions, which means U-Boot v2016.09+ will ignore it. On the other hand, older U-Boot versions won't care about this field (as it was "reserved" before), even if the new `sunxi-fel` assigns to it.

Issue #58, opened 2016-09-19
Please do a current release


StefanBruens:
Since the last release, a lot of changes have accumulated. Please evaluate if the current state has release quality and tag it.

Tnx.

far5893:
Quote, and add a --version flag to all commands

n1tehawk:
The current plan is to first integrate support for environment "overrides" (#60) and some smaller fixes. With those merged, we should be in a good state to tag a new release. I've also added a proposal to support version information (#62) for the various sunxi-tools utilities - comments and suggestions welcome.

Regards, NiteHawk

n1tehawk:
[sunxi-tools v1.4](https://github.com/linux-sunxi/sunxi-tools/releases/tag/v1.4) released

Note: From now on, programs should display version information in their usage help. For those binaries that do not output their usage by default (e.g. because they would process _stdin_), you may pass a `-?` option to get help, and thus version information.

Regards, NiteHawk

PR #62, opened 2016-10-22
Add support for version information 

n1tehawk:
See https://github.com/linux-sunxi/sunxi-tools/issues/58#issuecomment-255365876

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,10 @@
-sunxi-fexc
 bin2fex
 fex2bin
 sunxi-bootinfo
 sunxi-fel
-sunxi-pio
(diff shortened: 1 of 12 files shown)


n1tehawk:
I've been discussing this briefly with @ssvb on IRC, and we agreed that a "--version" option for `sunxi-fel` is prone to confusion with the "version" command (retrieving SoC information). I also dislike the fact that we have some programs that currently only support 'short' options (`-V`) and won't be able to support the 'long' `--version`.

The idea we came up is that we won't introduce new options, and instead will print the version information as part of the usage help. I'll rework this pull request accordingly.

PR #65, opened 2016-10-25
Fixed issue in progress=true in file_upload

howientc:
This fixes a problem introduced with the new progress functionality. progress was hard-coded as true in the call to aw_write_buffer() in the file_upload() function instead of paying attention to the progress callback passed in.

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -1492,7 +1492,9 @@ static unsigned int file_upload(libusb_device_handle *handle, size_t count,
 		void *buf = load_file(argv[i * 2 + 1], &size);
 		if (size > 0) {
 			uint32_t offset = strtoul(argv[i * 2], NULL, 0);
-			aw_write_buffer(handle, buf, offset, size, true);
+
+			/* fixed a problem with progress being passed in as true to aw_write_buffer */
+			aw_write_buffer(handle, buf, offset, size, (progress != NULL));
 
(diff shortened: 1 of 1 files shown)


fel.c
```diff
@@ -1492,7 +1492,9 @@ static unsigned int file_upload(libusb_device_handle *handle, size_t count,
 		void *buf = load_file(argv[i * 2 + 1], &size);
 		if (size > 0) {
 			uint32_t offset = strtoul(argv[i * 2], NULL, 0);
-			aw_write_buffer(handle, buf, offset, size, true);
+
+			/* fixed a problem with progress being passed in as true to aw_write_buffer */
```
n1tehawk: I think this comment should go into the commit message, not the actual code.

n1tehawk:
I'm curious: What actual issues did you perceive / encounter with the hardcoded `progress = true`? This parameter to `aw_write_buffer()` basically just indicates whether to support progress callbacks at all (true), or if it's desirable to suppress them (false). The current value shouldn't really be "harmful".

Nevertheless, the patch is useful - without a progress callback (set), there's no point in supporting the `progress_update()` call from within `usb_bulk_send()` - and when setting `progress = false` that routine is free to pick a slightly larger transfer window (AW_USB_MAX_BULK_SEND).

Regards, NiteHawk

n1tehawk:
Superseded by #66

PR #66, opened 2016-10-25
Upload progress fix2

howientc:
This pull request removes my improper "fixed issue" comment in the source and just contains the necessary fix itself.

The issue encountered was that by setting progress=true, when usb_bulk_send() is eventually called, progress==true has the side effect of setting the chunk size to 128K always:
Line 104:	size_t max_chunk = progress ? 128 * 1024 : AW_USB_MAX_BULK_SEND;

Regards,
Howie

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -1492,10 +1492,7 @@ static unsigned int file_upload(libusb_device_handle *handle, size_t count,
 		void *buf = load_file(argv[i * 2 + 1], &size);
 		if (size > 0) {
 			uint32_t offset = strtoul(argv[i * 2], NULL, 0);
-
-			/* fixed a problem with progress being passed in as true to aw_write_buffer */
 			aw_write_buffer(handle, buf, offset, size, (progress != NULL));
-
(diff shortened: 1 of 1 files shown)


n1tehawk:
Hi Howie!

I assume that you did this by mistake - but judging from the resulting commits, it was a bit pointless to create another branch and a new pull request for this. Preferably you would have appended (or possibly squashed and force-pushed your commits) to your existing branch, updating the previous PR at the same time.

For future contributions, please try to find meaningful commit messages. They allow people to understand the motivation behind your changes, and help to maintain a useful file/commit history. Additionally, the sunxi-tools repository also follows a policy where authors usually sign (`git commit -s`) their contributions.

The way it is, I think I'll have to pull in your requested change (plus a commit msg) via another branch and then discard this PR.

Regards, NiteHawk

howientc:
Hi Nitehawk,

Thanks for taking the time to set me straight. My git skills are lacking,
but improving. Next time, I'll get it right.

-Howie

n1tehawk:
Closed by merging #67

Issue #70, opened 2016-10-26
sunxi-tools 1.4 make fails when building meminfo


daym:
``` shell
$ make TARGET_TOOLS=meminfo
...cc -g -O0 -Wall -Wextra  -std=c99 -D_POSIX_C_SOURCE=200112L -D_BSD_SOURCE -D_DEFAULT_SOURCE -Iinclude/    meminfo.c Makefile common.h version.h   -o meminfo
....Makefile: file not recognized: File format not recognized
```

That is because we add Makefile, common.h and version.h as dependencies to all the tools (which I think is good).

However, the entry for sunxi-meminfo says:

``` Makefile
sunxi-meminfo: meminfo.c
        $(CROSS_COMPILE)gcc -g -O0 -Wall -static -o $@ $^
```

And $^ will pass the Makefile to the compiler - which will barf.

I'd pass $<

Same for sunxi-script_extractor and sunxi-fel and the generic sunxi-% rule.

daym:
See pull request #71

n1tehawk:
Oh... I've never thought of invoking `make` that way, overriding the *TARGET_TOOLS* variable. The original intent was that users would use `make sunxi-meminfo`, possibly with a custom *CROSS_COMPILE* setting, e.g.
```.sh
sunxi-tools # make -B sunxi-meminfo CROSS_COMPILE=armv7a-hardfloat-linux-gnueabi-
armv7a-hardfloat-linux-gnueabi-gcc -g -O0 -Wall -static -o sunxi-meminfo meminfo.c
```
This is also what README.md states.

Regards, NiteHawk

daym:
Well yeah, but that's what TARGET_TOOLS is supposed to be for, right? To list tools which only work on the target..

n1tehawk:
You're right - it took me a while to see all the implications. For those targets that have a single `.c` file dependency, the proposed change from `$^` to `$<` is a good suggestion.

However the longer I look at it, the more I get the impression that we're **lacking a dedicated rule for** `sunxi-pio`, meaning it gets built in the wrong way - namely using the native compiler of the host system instead of the cross-compiler aiming at the target.

Fixing that (by adding a proper rule) will also mean that me might "suddenly" see build failures for systems where no cross-compiler is available, because so far the *target-tools* is part of the default `make` target. I wonder if it would be preferable to make *tools* the default, and require users to explicitly specify `make all` if they also want the *target-tools*...

Regards, NiteHawk

[...]

PR #71, opened 2016-10-26
Fix build of sunxi-meminfo and sunxi-script_extractor and possible future programs

daym:
See issue #70

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ LIBUSB_CFLAGS = `pkg-config --cflags $(LIBUSB)`
 LIBUSB_LIBS = `pkg-config --libs $(LIBUSB)`
 
 sunxi-fel: fel.c fel-to-spl-thunk.h progress.c progress.h
-	$(CC) $(CFLAGS) $(LIBUSB_CFLAGS) $(LDFLAGS) -o $@ $(filter %.c,$^) $(LIBS) $(LIBUSB_LIBS)
+	$(CC) $(CFLAGS) $(LIBUSB_CFLAGS) $(LDFLAGS) -o $@ fel.c progress.c $(LIBS) $(LIBUSB_LIBS)
 
 sunxi-nand-part: nand-part-main.c nand-part.c nand-part-a10.h nand-part-a20.h
 	$(CC) $(CFLAGS) -c -o nand-part-main.o nand-part-main.c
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -111,7 +111,7 @@ sunxi-nand-part: nand-part-main.c nand-part.c nand-part-a10.h nand-part-a20.h
 	$(CC) $(LDFLAGS) -o $@ nand-part-main.o nand-part-a10.o nand-part-a20.o $(LIBS)
 
 sunxi-%: %.c
-	$(CC) $(CFLAGS) $(LDFLAGS) -o $@ $(filter %.c,$^) $(LIBS)
+	$(CC) $(CFLAGS) $(LDFLAGS) -o $@ $(filter %.c,$<) $(LIBS)
```
n1tehawk: Keeping `$filter()` but applying it to `$<` now doesn't really make sense, or? As it is, this change breaks compilation of `sunxi-fexc`.

Makefile
```diff
@@ -153,10 +153,10 @@ boot_head_sun5i.elf: boot_head.S boot_head.lds
 sunxi-bootinfo: bootinfo.c
 
 sunxi-meminfo: meminfo.c
-	$(CROSS_COMPILE)gcc -g -O0 -Wall -static -o $@ $^
+	$(CROSS_COMPILE)gcc -g -O0 -Wall -static -o $@ $<
 
 sunxi-script_extractor: script_extractor.c
-	$(CROSS_COMPILE)gcc -g -O0 -Wall -static -o $@ $^
+	$(CROSS_COMPILE)gcc -g -O0 -Wall -static -o $@ $<
```
n1tehawk: My impression is that we're missing an actual cross-compile rule for `sunxi-pio` here. I would favor introducing some common *TARGET_CFLAGS* and then use something like:
```Makefile
# target tools
TARGET_CFLAGS = -std=c99 $(DEFINES) -Iinclude/ -g -O0 -Wall -static
sunxi-meminfo: meminfo.c
	$(CROSS_COMPILE)gcc $(TARGET_CFLAGS) -o $@ $<
sunxi-pio: pio.c
	$(CROSS_COMPILE)gcc $(TARGET_CFLAGS) -o $@ $<
sunxi-script_extractor: script_extractor.c
	$(CROSS_COMPILE)gcc $(TARGET_CFLAGS) -o $@ $<
```

Makefile
daym: Yeah, sorry, it's too paranoid now. 

I don't really like these magical rules - I think it's much preferrable and less error-prone to have an explicit rule for each target file.

Makefile
daym: I agree.

Makefile
```diff
@@ -102,7 +102,7 @@ LIBUSB_CFLAGS = `pkg-config --cflags $(LIBUSB)`
 LIBUSB_LIBS = `pkg-config --libs $(LIBUSB)`
 
 sunxi-fel: fel.c fel-to-spl-thunk.h progress.c progress.h
-	$(CC) $(CFLAGS) $(LIBUSB_CFLAGS) $(LDFLAGS) -o $@ $(filter %.c,$^) $(LIBS) $(LIBUSB_LIBS)
+	$(CC) $(CFLAGS) $(LIBUSB_CFLAGS) $(LDFLAGS) -o $@ fel.c progress.c $(LIBS) $(LIBUSB_LIBS)
```
daym: You are right, it should be fine to use $(filter ) in this case - as it was in the beginning.

n1tehawk:
The PR was obsoleted by merging #73

Issue #78, opened 2016-11-08
Fix fel-copy.c, clarify its usage (or remove it completely)


n1tehawk:
`fel-copy.c` doesn't seem to be referenced/used anywhere. It looks like it's supposed to be ARM "scratch" code that could be uploaded and executed via FEL, to assist with arbitrary copy operations (of 32-bit words) within SoC memory.

However, the [current implementation](https://github.com/linux-sunxi/sunxi-tools/blob/1c3a6ca549dc76cf4725d28241923e20b1409a65/fel-copy.c#L40-L47) has notable problems, and doesn't match the "Parameters" description.

`b` starts pointing to `CONFIG_BASE` (0x2100); but with the post-increment logic applied as-is, it ends up pointing to 0x2108, which is also the "destination" where the subsequent loop will start copying data words to. For an actual destination address parameter (as per the description), at least one level of indirection is missing. `ptr` gets `b`'s address, meaning that the source pointer `a` is taken from (and later stored back to) 0x2100, not 0x2104. Something similar goes for the loop count `i`, which is taken from 0x2104, not 0x2108.

Either clean up this mess, or get rid of it. A possible alternative would be to have the scratch code in _fel.c_ and add a correspondig `copy` (or `copyl`) command to the _sunxi-fel_ utility.

n1tehawk:
Obsoleted by implementing `./sunxi-fel memmove` (#88).

PR #79, opened 2016-11-10
fel: Modify readl/writel functions, add "copyl"

n1tehawk:
The "copyl" functionality also obsoletes `fel-copy.c`, which has been removed.


ssvb:
Why call it "copyl"? As for the "readl" command, the "l" suffix can be interpreted either as "long" (32-bit) or "little endian".

n1tehawk:
I wanted it to be the same as "readl"/"writel" in the sense of 'long'. 'Little endian' doesn't apply, as this is not a copy between (USB) host and target device, but within SoC memory - so it's always 'native' endianness.

The 'long' is to hint that this command transfers 32-bit words, contrary to a 'string' (= byte-oriented) copy.

Can you suggest a better name?

ssvb:
If we want to have this functionality implemented, then I would rather have a more generic "memmove" command. With byte granularity for the buffer size. And also with a proper handling of the overlapped buffers.

n1tehawk:
My goal / "use case" initially was to eliminate the need for `fel-copy.c` (see #78).

I can also imagine that this copy function (once factored out into a FEL/USB "library") eventually gets used in a refreshed _sunxi-pio_, which would gain the ability to access the PIO registers more directly, instead of taking the "`fel-pio.bin` + file access" route. Accessing hardware registers is where the 'long' approach matters, see the source comment in 9a03f17.

ssvb:
I think that we can just drop ``fel-copy.c`` without bothering to provide any replacement. As for the ``sunxi-pio`` functionality, it would be best to have it just integrated as a high level command of the ``sunxi-fel`` tool.

PR #81, opened 2016-11-10
[Experimental] Add initial H5 support

n1tehawk:
So far, only basic FEL functions (retrieving BROM version and SID) and _uart0-helloworld_ functionality have been tested.

### DIFF
--- a/uart0-helloworld-sdboot.c
+++ b/uart0-helloworld-sdboot.c
@@ -138,6 +138,7 @@ enum sunxi_gpio_number {
 #define SUN5I_GPB_UART0         (2)
 #define SUN6I_GPH_UART0         (2)
 #define SUN8I_H3_GPA_UART0      (2)
+#define SUN50I_H5_GPA_UART0     (2)
 #define SUN50I_A64_GPB_UART0    (4)
 #define SUNXI_GPF_UART0         (4)
 
@@ -288,6 +289,11 @@ int soc_is_h3(void)
 	return soc_id == 0x1680;
 }
 
+int soc_is_h5(void)
+{
+	return soc_id == 0x1718;
+}
+
 /*****************************************************************************
  * UART is mostly the same on A10/A13/A20/A31/H3/A64, except that newer SoCs *
  * have changed the APB numbering scheme (A10/A13/A20 used to have APB0 and  *
@@ -347,6 +353,10 @@ void gpio_init(void)
 	} else if (soc_is_h3()) {
 		sunxi_gpio_set_cfgpin(SUNXI_GPA(4), SUN8I_H3_GPA_UART0);
 		sunxi_gpio_set_cfgpin(SUNXI_GPA(5), SUN8I_H3_GPA_UART0);
+		sunxi_gpio_set_pull(SUNXI_GPA(5), SUNXI_GPIO_PULL_UP);
+	} else if (soc_is_h5()) {
+		sunxi_gpio_set_cfgpin(SUNXI_GPA(4), SUN50I_H5_GPA_UART0);
+		sunxi_gpio_set_cfgpin(SUNXI_GPA(5), SUN50I_H5_GPA_UART0);
 		sunxi_gpio_set_pull(SUNXI_GPA(5), SUNXI_GPIO_PULL_UP);
 	} else {
 		/* Unknown SoC */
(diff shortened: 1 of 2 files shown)


uart0-helloworld-sdboot.c
```diff
@@ -247,51 +247,28 @@ void soc_detection_init(void)
...
-{
-	return soc_id == 0x1633;
-}
+#define soc_is_a10	(soc_id == 0x1623)
+#define soc_is_a20	(soc_id == 0x1651)
+#define soc_is_a31	(soc_id == 0x1633)
+#define soc_is_a80	(soc_id == 0x1639)
+#define soc_is_a64	(soc_id == 0x1689)
+#define soc_is_h3	(soc_id == 0x1680)
+#define soc_is_h5	(soc_id == 0x1718)
```
ssvb: If we convert these to macros, then it would be preferable to have them uppercase, adhering to a common C coding style convention.

ssvb:
Except for the refactoring part, this code exactly matches the changes that I have tested locally myself. Does @Icenowy still [https://irclog.whitequark.org/linux-sunxi/2016-11-10#18141638; get zero as a SID value] on her board?

Icenowy:
2016年11月11日 09:33于 Siarhei Siamashka <notifications@github.com>写道：
>
> Except for the refactoring part, this code exactly matches the changes that I have tested locally myself. Does @Icenowy still [https://irclog.whitequark.org/linux-sunxi/2016-11-10#18141638; get zero as a SID value] on her board?

I got something not zero.

Thanks!

ssvb:
> I got something not zero.

OK, then it seems like everything is fine now.

n1tehawk:
Okay, I'll strip the refactoring part for now - It doesn't really fit in here.

@ssvb Maybe we could find a compromise by declaring 'parametrized' macros like
`#define soc_is_a20()	(soc_id == 0x1651)`
which causes the compiler to complain when used without the brackets. Admittely it's even more covering up the macro nature that way...

ssvb:
@n1tehawk Dropping the refactoring patch is fine for me. But if declaring 'parametrized' macros will make the compiler safeguard us against writing wrong code constructs, then this is actually ideal. And we don't need to convert names to uppercase then, because the C library has some function-alike macros too.

Icenowy:
2016年11月11日 10:08于 Siarhei Siamashka <notifications@github.com>写道：
>
> @n1tehawk Dropping the refactoring patch is fine for me. But if declaring 'parametrized' macros will make the compiler safeguard us against writing wrong code constructs, then this is actually ideal. And we don't need to convert names to uppercase then, because the C library has some function-alike macros too.

P.S. I tried to boot a Boot0 via `fel spl boot0.img`, Boot0 is booted, however, after loading next stage bootloader from SD, the system stucks...

ssvb:
> P.S. I tried to boot a Boot0 via `fel spl boot0.img`, Boot0 is booted, however, after loading next stage bootloader from SD, the system stucks...

This is fine. The boot0 bootloader just does not try to return back to FEL and we can't really do anything about this. And it is already mentioned at the http://linux-sunxi.org/FEL/USBBoot#Testing page as ``"Please note that you may get error messages from the 'sunxi-fel' tool while doing this. This is normal and expected (boot0 does not return control back to the FEL code in BROM but instead does its usual booting of the system from the SD card)."``

Icenowy:
2016年11月11日 10:16于 Siarhei Siamashka <notifications@github.com>写道：
>>
>> P.S. I tried to boot a Boot0 via fel spl boot0.img, Boot0 is booted, however, after loading next stage bootloader from SD, the system stucks...
>
> This is fine. The boot0 bootloader just does not try to return back to FEL and we can't really do anything about this. And it is already mentioned at the http://linux-sunxi.org/FEL/USBBoot#Testing page as "Please note that you may get error messages from the 'sunxi-fel' tool while doing this. This is normal and expected (boot0 does not return control back to the FEL code in BROM but instead does its usual booting of the system from the SD card)."

But it cannot also "does it usual booting of the system from the SD card".

ssvb:
> But it cannot also "does it usual booting of the system from the SD card".

Hmm, it worked like this with the older SoCs. You enter the FEL mode, then insert the SD card (so that the boot0 can find the second stage bootloader on it), then try to do FEL boot of the boot0.

You may try to debug this issue if you are interested. Comparing the serial console logs for the FEL boot of the boot0 with the normal SD card boot of the boot0 may be the first step.

PR #82, opened 2016-11-13
Refactoring / minor changes

n1tehawk:


README.md
```diff
@@ -8,7 +8,7 @@ Copyright (C) 2012  Alejandro Mery <amery@geeks.cl>
 [this link](https://github.com/linux-sunxi/sunxi-tools/contributors)
 or use the command `git shortlog -se --no-merges`.
 
-Tools to help hacking Allwinner A10 (aka sun4i) based devices and
+Command line utilities to work with Allwinner A10 (aka sun4i) based devices and
```
ssvb: While at it, we may remove this emphasis on Allwinner A10 and sun4i.

ssvb:
Maybe we don't need to mention Nintendo at all in the commit message. Just say that the word hacking is ambiguous because it has more than one meanings nowadays. For example, http://www.thefreedictionary.com/hacking defines "hacking" as ``"a. To write or refine computer programs skillfully"`` and ``" To use one's skill in computer programming to gain illegal or unauthorized access to a file or network"``. And http://dictionary.cambridge.org/dictionary/english/hacking only has negative connotations.

uart0-helloworld-sdboot.c
```diff
@@ -247,51 +247,28 @@ void soc_detection_init(void)
...
-{
-	return soc_id == 0x1633;
-}
+#define soc_is_a10()	(soc_id == 0x1623)
+#define soc_is_a20()	(soc_id == 0x1651)
+#define soc_is_a31()	(soc_id == 0x1633)
+#define soc_is_a80()	(soc_id == 0x1639)
+#define soc_is_a64()	(soc_id == 0x1689)
+#define soc_is_h3()	(soc_id == 0x1680)
+#define soc_is_h5()	(soc_id == 0x1718)
```
ssvb: Please mention in the commit message that using defines is safer because the compiler does not allow to interpret them as function pointers (if accidentally used without brackets). And we have less lines of code as a bonus.

n1tehawk: Okay. Actually the preprocessor will ignore the macro in that case, meaning the compiler will end up with a "call to undefined function in ..." - but the net effect / result is the same, of course.

n1tehawk: Well, getting rid of the function body was my primary motivation here.

`inline` shouldn't make much of a difference here - even without it, the compiler will complain (as long as warnings are enabled):
```
uart0-helloworld-sdboot.c:337:13: warning: the address of ‘soc_is_h5’ will always evaluate as ‘true’ [-Waddress]
  } else if (soc_is_h5) {
             ^
```

ssvb:
Thanks! The first two patches look good to me now.

PR #85, opened 2016-11-29
fel: Add SID register address for A80

wens:
The SID block in the A80 is at 0x01c0e000, with the e-fuses we care
about at offset 0x200 within the block.

Signed-off-by: Chen-Yu Tsai <wens@csie.org>


n1tehawk:
Merged. Thanks!

PR #88, opened 2016-12-17
fel: Implement memory copy operations and "memmove" command

n1tehawk:
This introduces new ARM thunk code implementing a _memmove(3)_-compatible approach, which means that the `fel_memmove()` function is able to handle overlap and 'unaligned' access gracefully.

Also: With the corresponding `memmove` command in _sunxi-fel_, the old `fel-copy.c` is now obsolete and thus removed.

(EDIT: This is a follow-up to #79)


ssvb:
The code looks fine to me. Still the question is whether there any real practical use for it.

n1tehawk:
Agreed. That's why I'm still hesitant to merge it. The main idea was getting rid of `fel-copy.c`, while still providing a functional equivalent - whether needed or not.

Admittedly even the use case of "remote" PIO register access via FEL would probably already be covered fine using the `fel_readl_n()`/`fel_writel_n()` functions.

n1tehawk:
I'm pretty sure that the `fel-gpio` script has fallen into disuse (if if ever was anything else but 'hacking' to begin with), but it still references some memory-shoveling that might get replaced by "memmove": [fel-gpio#L32](https://github.com/linux-sunxi/sunxi-tools/blob/f1cb74d828731c5bca92429af8198fae5f689651/fel-gpio#L32)  (and #L40-L41).

PR #90, opened 2016-12-19
fel: workaround H3 SID issue

Icenowy:
H3 SID controller has some bug, that makes the initial value at
0x01c14200 wrong.

This commit workarounds this bug by reading them with register access
first.

Signed-off-by: Icenowy Zheng <icenowy@aosc.xyz>

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -276,11 +276,49 @@ void fel_writel(feldev_handle *dev, uint32_t addr, uint32_t val)
 	fel_writel_n(dev, addr, &val, 1);
 }
 
+#define EFUSE_PRCTL 0x40
+#define EFUSE_RDKEY 0x60
+
+#define EFUSE_OP_LOCK 0xAC
+
+#define EFUSE_SRAM_BASE 0x200
+
+uint32_t aw_fel_efuse_read(feldev_handle *dev, uint32_t offset)
+{
+	soc_info_t *soc_info = dev->soc_info;
+	uint32_t reg_val;
+
+	uint32_t efuse_base = soc_info->sid_addr - EFUSE_SRAM_BASE;
+
+	reg_val = fel_readl(dev, efuse_base + EFUSE_PRCTL);
+	reg_val &= ~(((0x1ff) << 16) | 0x3);
+	reg_val |= (offset << 16);
+	fel_writel(dev, efuse_base + EFUSE_PRCTL, reg_val);
+
+	reg_val &= ~(((0xff) << 8) | 0x3);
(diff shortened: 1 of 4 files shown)


n1tehawk:
For the rationale behind this, see https://groups.google.com/forum/#!topic/linux-sunxi/ynyIP8c61Qs. I've also added some additional findings in my comment at https://groups.google.com/d/msg/linux-sunxi/ynyIP8c61Qs/v33UFC2LEgAJ.

One thing I do not like too much in the FEL patch is that it requires a somewhat excessive use of the readl/writel functions over USB (for 'bit-banging' the SID_PRCTL register). Maybe this should be transformed to a piece of thunk code that can be executed on the target device - I'll give that a try when I find some time.

Regards, NiteHawk

Icenowy:
2016年12月20日 19:33于 NiteHawk <notifications@github.com>写道：
>
> For the rationale behind this, see https://groups.google.com/forum/#!topic/linux-sunxi/ynyIP8c61Qs. I've also added some additional findings in my comment at https://groups.google.com/d/msg/linux-sunxi/ynyIP8c61Qs/v33UFC2LEgAJ.
>
> One thing I do not like too much in the FEL patch is that it requires a somewhat excessive use of the readl/writel functions over USB (for 'bit-banging' the SID_PRCTL register). Maybe this should be transformed to a piece of thunk code that can be executed on the target device - I'll give that a try when I find some time.

I cannot write ARM assembly...

Sorry for this.

n1tehawk:
Superseded by merging #91.

ssvb:
Looks like a good fix to me, great job. We could also add a new helper function ```fel_clrsetbits_le32``` (similar to the one used in U-Boot) in addition to ```fel_readl``` and ```fel_writel```. It can make such code cleaner and reduce the number of round-trips a little bit.

ssvb:
> One thing I do not like too much in the FEL patch is that it requires a somewhat excessive use of the readl/writel functions over USB (for 'bit-banging' the SID_PRCTL register). Maybe this should be transformed to a piece of thunk code that can be executed on the target device - I'll give that a try when I find some time.

There is not much point doing this, unless we have a real performance problem at our hands. There are different programming languages such as ```Assembly```, ```C``` and ```Python/Ruby/Lua``` for a reason. Assembly code provides absolute control where necessary and allows best performance, but is difficult to implement and maintain. C loses a bit of control, but still allows low level stuff and is much easier than Assembly. Then we have scripting languages, which make the code much easier to write and maintain, but sacrifice some runtime performance (actually often really a lot). Scripting languages have their niche because for some tasks it really does not matter if the program does the job in 0.1 or in 0.001 second.

PR #91, opened 2016-12-21
SID fix / H3 workaround

n1tehawk:
This pull request extends on #90, and provides ARM thunk code for register-based retrieval of the SID (root) key.

I've tested this on A20 (Banana Pi M1), H3 (Orange Pi PC) and A64 (Pine64+).


ssvb:
I tried to benchmark the performance of "sunxi-fel sid". The original fix from @Icenowy needs ~0.07s per one SID retrieval:

<pre>
$ time sh print-100-sids.sh > /dev/null

real	0m7.162s
user	0m0.280s
sys	0m0.740s
</pre>

The optimization from @n1tehawk needs 0.009s (~8 times faster):

<pre>
$ time sh print-100-sids.sh > /dev/null

real	0m0.911s
user	0m0.010s
sys	0m0.010s
</pre>

n1tehawk:
>If we want some diagnostic capabilities without overcomplicating the code too much, then we could just read and compare the SID value before and after the fixup (so that we know whether activating the workaround via the ```.sid_fix``` field was really necessary). And additional user visible command just for that may be an overkill.

Yes, but how would you "activate" this? Currently there's no "debug" switch (or build) available. At least for the time being, I wanted to provide a simple way of enabling users to check for this particular issue (if they suspect that other SoCs might be affected too).

Regards, NiteHawk

PR #93, opened 2017-01-03
Improve description for uart0-helloworld-sdboot

plaes:
Signed-off-by: Priit Laes <plaes@plaes.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,12 +51,13 @@ to allow GPIO manipulations via FEL
 ARM native sdcard bootloader forcing the device into FEL mode
 
 ### uart0-helloworld-sdboot
-ARM native sdcard bootloader, which is only printing a short "hello"
-message to the UART0 serial console. Because it relies on runtime
-SoC type detection, this single image is bootable on a wide range of
-Allwinner devices and can be used for testing. Additionally, it may
-serve as a template/example for developing simple bare metal code
-(LED blinking and other similar GPIO related things).
+ARM native sdcard bootloader, which only prints a short "Hello from <soc>"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,12 +51,13 @@ to allow GPIO manipulations via FEL
...
 
 ### uart0-helloworld-sdboot
-ARM native sdcard bootloader, which is only printing a short "hello"
-message to the UART0 serial console. Because it relies on runtime
-SoC type detection, this single image is bootable on a wide range of
-Allwinner devices and can be used for testing. Additionally, it may
-serve as a template/example for developing simple bare metal code
-(LED blinking and other similar GPIO related things).
+ARM native sdcard bootloader, which only prints a short "Hello from <soc>"
+message with SoC variant to the UART0 serial console and returns to FEL mode.
```
ssvb: This is not exactly accurate. When booting from the SD card, it enters an infinite loop after printing the message. Though updating it to actually jump to FEL instead of this infinite loop would be probably a useful feature.

plaes: Actually, I wouldn't bother. When flashing via `sunxi-fel`, the looping can be done on the host machine.

I was only confused, because I wanted to use that message to debug whether I got my RX/TX correct, but I didn't see the message continuously written.

ssvb: Do you mean that you connect your UART cable (or start your terminal application) with a bit of delay after powering up the board and don't see this message?

plaes: Yes... a bit of a delay and as usual - RX/TX swapped ;)

plaes:
OK, let's close this one.

PR #94, opened 2017-01-26
Move thunk code to a dedicated subdirectory, implement fel_clrsetbits_le32()

n1tehawk:


n1tehawk:
Implement `fel_clrsetbits_le32()`, as suggested in #90.

n1tehawk:
Will reorder / squash some commits in preparation for merge...

ssvb:
Looks good in general. Just some minor nitpicks:
* The choice of name "thunk" may be not very good. Wikipedia defines https://en.wikipedia.org/wiki/Thunk more or less as an adapter between something (such as the calling conventions translation or other things).
* The find invocation in the makefile needs to have ```-maxdepth 1``` option, because otherwise the search is recursive. I had a wrong directory in my PATH (pointing to the android toolchain root directory rather than its "bin" subdirectory) and got some really weird results :)
* Regarding the license. Now README.md says that no explicit notice means the GPL license. Maybe we want to automatically add some sort of a boilerplate notice to the *.h files, stating that they are automatically generated and the actual license needs to be checked in the corresponding *.S files? As for the license choice for the ```fel-to-spl-thunk.S``` file, I picked the MIT license just to make it clear that "sunxi-fel spl" is allowed to work with proprietary bootloaders too (so that the remotely executed bootloader is not GPL poisoned by the remotely executed thunk code).

PR #95, opened 2017-02-07
Misc small build nitpicks

plaes:


n1tehawk:
Proposed changes look good to me.

As this is already touching the _find_ invocation, maybe we should also integrate a `-maxdepth 0` limit similar to what @ssvb proposed [here](https://github.com/linux-sunxi/sunxi-tools/pull/94#issuecomment-278117675), so that `find` only takes a look at the actual directories contained within `PATH`, excluding any subfolder.

We should also try to keep the `CROSS_COMPILE` detection in _thunks/Makefile_ identical. Maybe this should go into a shell script (_find-arm-gcc.sh_) that both Makefiles could make use of?

Regards, NiteHawk

n1tehawk:
The current _PATH_ handling also has issues with directory names containing spaces. I think we should better use something like
```sh
#
# Try to locate suitable ARM cross compilers available via $PATH
# If any are found, this function will output them as a TAB-delimited list
#
scan_path () {
echo ${PATH} | sed -e 's/:/\n/g' | while read path; do
	find "$path" -maxdepth 1 -executable -name 'arm*-gcc' -printf '%f\t' 2>/dev/null
done
}

# Use only the first field from result, and convert it to a toolchain prefix
scan_path | cut -f 1 | sed -e 's/-gcc/-/'
```

plaes:
Don't know if it is bash-specific, but we can use the "magic" `IFS` variable (internal field separator):
```
scan_path () {
        IFS=":"
        for path in $PATH; do
                find "$path" -maxdepth 1 -executable -name 'arm*-gcc' -printf '%f\t' 2>/dev/null
        done
}

# Use only the first field from result, and convert it to a toolchain prefix
scan_path | cut -f 1 | sed -e 's/-gcc/-/'
```

n1tehawk:
Good point. To my knowledge `IFS` is POSIX-compliant, and should work with other shells too.

PR #98, opened 2017-02-13
fel: Try to detect VM environments and print a warning for those

n1tehawk:
We try to detect virtual machines by parsing the vendor IDs of all USB devices, and will print a warning message if such an environment is encountered.

### DIFF
--- a/fel_lib.c
+++ b/fel_lib.c
@@ -820,3 +820,47 @@ feldev_list_entry *list_fel_devices(size_t *count)
 	if (count) *count = devices;
 	return list;
 }
+
+/*
+ * Check for virtual machine (USB emulation). The FEL protocol is rather
+ * sensitive to errors and timing issues, so such environments are known
+ * to NOT work well with sunxi-fel.
+ *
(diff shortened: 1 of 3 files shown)


fel_lib.c
```diff
@@ -820,3 +820,47 @@ feldev_list_entry *list_fel_devices(size_t *count)
...
+ */
+const char *fel_check_vm(void) {
+	const char *result = NULL;
+
+	ssize_t rc;
+	libusb_context *ctx;
+	libusb_device **usb;
+	struct libusb_device_descriptor desc;
+
+	libusb_init(&ctx);
```
ssvb: Do we need to check for error code from ``libusb_init()`` too? http://libusb.sourceforge.net/api-1.0/group__lib.html#ga9517c37281bba0b51cc62eba728be48b

kotc:
Tested-By: @kotc

tested in virtualbox/debian64, vm was detected and warning printed ok

ssvb:
Should we merge this as-is or try to implement @apritzel's suggestion?

ssvb:
Okay, I will push this to the master branch in a few days.

PR #99, opened 2017-02-28
fel: SMC workaround for the Allwinner SoCs with the secure bit set in eFUSE

ssvb:
A one time programmable secure bit in the eFUSE radically changes the boot behaviour on some SoCs (H3/A64/H64). A new bootloader header format https://linux-sunxi.org/TOC0 replaces the old eGON.BT0. Also the FEL handler starts as non-secure and has no access to a lot of resources (SID, SRAM A2, etc.). Fortunately a workaround to switch FEL into the secure mode exists and these patches implement it.

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -403,6 +403,39 @@ void aw_set_sctlr(feldev_handle *dev, soc_info_t *soc_info,
 }
 
 /*
+ * Issue a "smc #0" instruction. This brings a SoC booted in "secure boot"
+ * state from the default non-secure FEL into secure FEL.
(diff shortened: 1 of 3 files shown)


ssvb:
If there are no objections, I'm going to merge this pull request in a few days. Thanks.

hno:
Looks fine. But I cannot test as I have no secure device.

ssvb:
If you have a Pine64 board or Orange Pi PC, then you can switch it to secure. We can add the exact instructions to the wiki (at the TOC0 page), but you can't undo this once the secure bit is set.

ssvb:
Merged.

PR #102, opened 2017-05-18
fel: enable support for v2 SPL

Icenowy:
The version 2 of SPL added the possibility to add a device tree name in
the header, with adding some pad and using a reserved word.

As FEL boot currently doesn't need the device tree name, directly raise
the maximum supported version number to 2.

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>


plaes:
Tested-by: Priit Laes <plaes@plaes.org>

A20 - Cubietruck.

PR #103, opened 2017-06-07
Added support for '-' and '/' characters.

RavRabbit:
Newer fex files like the one [here](https://github.com/OrangePiLibra/OrangePiH5_external/blob/master/sys_config/OrangePiH5_PC2_sys_config.fex) cannot be converted to bin.

> invalid character at 4.
There are fex files with "-" and "/" characters in their section header or key. These characters aren't supported with fex2bin.

### DIFF
--- a/script_bin.c
+++ b/script_bin.c
@@ -242,7 +242,7 @@ static int decompile_section(void *bin, size_t bin_size,
 		words	= (entry->pattern >>  0) & 0xffff;
 
 		for (char *p = entry->name; *p; p++)
-			if (!(isalnum(*p) || *p == '_')) {
+			if (!(isalnum(*p) || *p == '_' || *p == '-' || *p == '/')) {
 				pr_info("Warning: Malformed entry key \"%s\"\n",
 					entry->name);
 				break;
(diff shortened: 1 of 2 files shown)


ssvb:
Thanks for the patches.

The '-' and '/'' support patch looks mostly good. Just a few nitpicks:
* the commit summary would look better with "fex:" prefix
* the commit message could provide a detailed description of the problem (basically, take the text from your first comment in this pull request).
* we have a test suite in sunxi-tools, and it would be nice to extend it to cover these new section names and keys too.

Do you have some step by step instructions explaining how to use FEX files on Allwinner H5? If I remember correctly, it's a 64-bit SoC and Allwinner tried to switch to DTS at least for A64. Did H5 move back to FEX?

Regarding the B288 patch, it is independent from the FEX changes and belongs to a separate pull request. Did you test it on some hardware? Can you try to add B288 support to uart0-helloworld too?

plaes:
I'm not 100% sure whether we would like to allow `/` character in the property names. For sections ( aka '[foo/bar]`, it is ok.

Erron-Black (approved):

apritzel:
This is superseded by the very similar https://github.com/linux-sunxi/sunxi-tools/pull/171/commits/edec4d2f6c071c5aadbd7e0075abdafb4031e6d7, which fixes the concern that @plaes mentioned above. As this one is merged, this PR can be closed.

PR #107, opened 2017-09-10
fel: SPI flash support

apritzel:
Hi,

this is a rebased version of @ssvb's original SPI flash support patches(63dfa9fed5c87e378ec9b42ce93784f4d96053c8). I just adapted his patches to the upstream changes since (mostly libusb_device_handle -> feldev_handle and filling the gaps), also dropped patch 2 (which introduced a header file), with I consider superseded by the fel_lib refactoring.
To be honest I am overly happy with this series (do we really need a Ruby script?), but it's far too useful to let it bit-rot.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,8 +130,9 @@ HOST_CFLAGS = $(DEFAULT_CFLAGS) $(CFLAGS)
 PROGRESS := progress.c progress.h
 SOC_INFO := soc_info.c soc_info.h
 FEL_LIB  := fel_lib.c fel_lib.h
+SPI_FLASH:= fel-spiflash.c fel-spiflash.h fel-remotefunc-spi-data-transfer.h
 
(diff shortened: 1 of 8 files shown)


ssvb:
The ruby script is needed if we want to recompile the code, which is running on the device side. Rather than carrying this blob in the " fel-remotefunc-spi-data-transfer.h" header file.

I'll check what can be done to update and merge this code.

apritzel:
I see the point of this, and I think it paves the road to future extensions, for instance (e)MMC access. Also as the Ruby environment is not required for compiling sunxi-fel, that's not a problem, I think.

On a related matter: Shall there be some (optional) test make target to trigger this re-creation? At the moment no-one will probably realise when there is a bug in the device side source and when some refactoring requires adjustment. I stumbled upon this problem when reviewing @Icenowy's version.

karlp:
would be nice for this to land?  It's referred to in the u-boot readmes, but "doesn't exist" yet :)

apritzel:
Many thanks to @wens for merging this!
Closing.

PR #108, opened 2017-11-06
Uboot crc

mripard:
Some recent changes to U-Boot added some padding to the U-Boot images, which improperly trigger the check on the size consistency.

Fix this by checking the CRC instead, and make a bunch of improvements along the way.


ssvb:
Looks good to me. And we can always get rid of the zlib dependency later.

wens:
Any other comments on this? Otherwise this is going tomorrow.

PR #109, opened 2017-11-22
Allwinner H6 support

Icenowy:


wens:
Can you split the uart0-helloworld-sdboot patch into two?
The first one should convert the *_BASE macro usage to the static global variables *_base.
The commit should state why this is being done, namely because the H6 has a different memory map.
The second will actually add H6 support.

uart0-helloworld-sdboot.c
```diff
@@ -266,8 +278,16 @@ void soc_detection_init(void)
 	if (((midr >> 4) & 0xFFF) == 0xC0F) {
 		soc_id = 0x1639; /* ARM Cortex-A15, so likely Allwinner A80 */
 	} else {
+		/*
+		 * The address of VER_REG in SoCs other than H6 is in the
+		 * VP9 part of H6, which is not powered up when booting.
+		 */
```
Andre-ARM: So this may be true for now, but still I would like to avoid **writing** somewhere in memory first.
I had the following idea:

    /*
     * Most SoCs have the SRAM controller at 0x01c00000, but the
     * H6 is special. So let's probe for that first. But before
     * writing into some random memory, let's first try to read
     * something from which we know the content of: the
     * GIC IIDR register.
     */
    reg = readl(0x03021004);        /* GICD_IIDR in H6 memory map*/
    if (reg & 0xfff == 0x43b)           /* GICv2 */
        reg = SUNXI_NEW_VER_REG;
    else
        reg = SUNXI_VER_REG;

So this **reads** from some area which has nothing on older SoCs, but returns a well known value on the H6.
This is untested yet.

soc_info.c
```diff
@@ -94,6 +94,18 @@ sram_swap_buffers a80_sram_swap_buffers[] = {
...
 
+/*
+ * H6 has 32KiB of SRAM A at 0x20000 and a large SRAM C at 0x28000. SRAM A
+ * and SRAM C reside in the address space back-to-back without any gaps, thus
+ * representing a singe large contiguous area. Everything is the same as on
+ * A10/A13/A20, but just shifted by 0x20000.
+ */
+sram_swap_buffers h6_sram_swap_buffers[] = {
+	{ .buf1 = 0x21C00, .buf2 = 0x2A400, .size = 0x0400 },
+	{ .buf1 = 0x25C00, .buf2 = 0x2A800, .size = 0x2400 },
```
Andre-ARM: Is there any reason you omitted the third line from A20/A64 ("something important")? I would expect those lines to be the same as for the A64, but just 0x10000 added to each **address** (so size stays at 0x1400 for the second buffer, for instance).

uart0-helloworld-sdboot.c
```diff
@@ -150,6 +156,8 @@ enum sunxi_gpio_number {
 
 static u32 pio_base;
 
+static u32 pio_base;
```
ssvb: A duplicate line?

ssvb:
Could you please add an entry about H6 to http://linux-sunxi.org/FEL/USBBoot#SoC_support_status ? Also what is the current status of ``fel-sdboot.S`` the on H6? Does it work correctly?

ssvb:
BTW, this new H6 SoC with SRAM A1 at the new address 0x20000 is a nice demonstration why I did not like the introduction of the CONFIG_SUNXI_HIGH_SRAM boolean option in U-Boot : https://irclog.whitequark.org/linux-sunxi/2016-11-24#18253072
So much for "only two alternatives" :)

Icenowy:
@ssvb fel-sdboot works on H6 as TL Lim tested.

In U-Boot really it's a problem... maybe I should then rewrite parts about SRAM A1 address.

Icenowy:
duplicated pio_base fixed.

Icenowy:
@ssvb could you check this PR again?

Andre-ARM:
Icenowy,

can you please fix the GIC sequence I managed to wrongly propose in my
review comment on the uart0-helloworld-sdboot.c? It should read:
...
	reg = readl(0x03021008);
	if ((reg & 0xfff) == 0x43b) /* Found GICv2 here, so it's a H6 */
		reg = H6_VER_REG;
	else
		reg = VER_REG;
....

Cheers,
Andre.

Icenowy:
@ssvb could you check this PR again?

zorgzerg:
Guys, please! Very need support H6

apritzel:
@Icenowy, @wens merged this (many thanks for that!), can you please check that it works for you and then close this PR? Thanks!

Icenowy:
Test passed. Will close the PR.

Issue #104, opened 2017-07-07
Boot info not passed via Sunxi SPL


pascallj:
Currently in the process of getting mainline U-boot and kernel ready for the Topwise A721. When trying the FEL USBboot, I noticed that when uploading `boot.scr` using `sunxi-fel`, the `fel_scriptaddr` does not get set in the SPL header and `bootcmd_fel` fails.

The `sunxi-fel` tool fails with:
```
sunxi SPL version mismatch: found 0x02 > maximum supported 0x01
You need a more recent version of this (sunxi-tools) fel utility.
```

However when manually increasing `SPL_MAX_VERSION` in `fel.c` to `2` the tool works as it should and I get

```
Passing boot info via sunxi SPL: script address = 0x43100000, uEnv length = 0
```
and U-boot loads the script automatically.

What is the deal here? I assume increasing the `SPL_MAX_VERSION` is not an correct solution, but what is?

pascallj:
Maybe it's useful to mention that U-boot changed the `SPL_HEADER_VERSION` to `2` in [this commit](http://git.denx.de/?p=u-boot.git;a=commitdiff;h=7f0ef5a945a0a9cef773c2b05b157d2fef3ea7ad).

Seems like there were some noticeable changes in version 2 as well, so changing just the version number is probably not a solution. 
As I am just getting started with this project, I guess I don't have the skills to update the tool. Maybe someone can?

ssvb:
Thanks for reporting! This is indeed quite a nasty problem and we unfortunately dropped the ball earlier. Here are the links to older mailing list discussions:

https://groups.google.com/d/msg/linux-sunxi/XuwQF_cK2jA/BPw5F_q-AgAJ
https://lists.denx.de/pipermail/u-boot/2015-September/228727.html

Basically, the SPL version header bump in U-Boot should not have caused any disruptions for the existing users if we had done it right. But unfortunately right now all the sunxi-tools packages in various linux distributions will have to be updated because of this bug. I'll try to fix this ASAP and we will need to tag a new release too.

ssvb:
Yes, just changing SPL_HEADER_VERSION to 2 is not good enough, because this does not fix the root cause of the problem. And we can't play a catch-up game.

pascallj:
Well you basically identified my problem back in 2015:
> Then we can have the users trying to follow some
tutorial and expecting their boot.scr to be picked up by U-Boot. And
[...]

PR #114, opened 2018-06-29
v1.5-rc1

### LINKED ISSUE #105: int main() without return in pio.c
got the following:

vendor/tvd/packages/sunxi-tools/pio.c: In function 'main':
vendor/tvd/packages/sunxi-tools/pio.c:430:1: error: control reaches end of non-void function [-Werror=return-type]
cc1: some warnings being treated as errors

declaring int main() should always include final return X

### LINKED ISSUE #106: typo in pio.c: ouput

### LINKED ISSUE #111: fel: relax SCTLR check
On a H2+ (Orange Pi Zero), if I enter fel mode by entering "go 0xFFFF0020" on u-boot cmd line, then sunxi-fel fails with this error: 

"Unexpected SCTLR (00C5287A)"

This does not happen when I enter fel mode directly on power-on (no sd card for eg).

I've attempted to remove this check and sunxi-fel can work fine without it (tested uboot command successfully).

I suggest SCTLR check is relaxed.

apritzel:
This PR gathers all outstanding patches that should make it into a new release. It contains the fixes I posted some days ago, the H6 support from @Icenowy (with two small fixes), and the SPI flash support from @ssvb, rebased and adapted to mainline. On top the "semantic versioning" fix, which would make sunxi-fel more robust against future additions to the SPL header and should avoid those nasty breakages we saw in the past (c8ada3849e9b).
Happy to have a discussion about which patch should go in, but most patches here have been around for a while and should really be merged now, since people keep asking for it.
Fixes #106. Fixes #111. Fixes #105.

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -243,6 +243,111 @@ void aw_fel_fill(feldev_handle *dev, uint32_t offset, size_t size, unsigned char
 	}
 }
 
+/*
+ * Upload a function (implemented in native ARM code) to the device and
+ * prepare for executing it. Use a subset of 32-bit ARM AAPCS calling
+ * conventions: all arguments are integer 32-bit values, and an optional
+ * return value is a 32-bit integer too. The function code needs to be
+ * compiled in the ARM mode (Thumb2 is not supported), it also must be
+ * a position independent leaf function (have no calls to anything else)
+ * and have no references to any global variables.
+ *
+ * 'stack_size'    - the required stack size for the function (can be
+ *                   calculated using the '-fstack-usage' GCC option)
+ * 'arm_code'      - a pointer to the memory buffer with the function code
+ * 'arm_code_size' - the size of the function code
+ * 'num_args'      - the number of 32-bit function arguments
+ * 'args'          - an array with the function argument values
+ *
+ * Note: once uploaded, the function can be executed multiple times with
+ *       exactly the same arguments. If some internal state needs to be
+ *       updated between function calls, then it's best to pass a pointer
+ *       to some state structure located elsewhere in SRAM as one of the
+ *       function arguments.
+ */
+
+bool aw_fel_remotefunc_prepare(feldev_handle *dev,
+			       size_t                stack_size,
+			       void                 *arm_code,
+			       size_t                arm_code_size,
+			       size_t                num_args,
+			       uint32_t             *args)
+{
+	size_t idx, i;
+	size_t tmp_buf_size;
+	soc_info_t *soc_info = dev->soc_info;
+	uint32_t *tmp_buf;
+	uint32_t new_sp, num_args_on_stack = (num_args <= 4 ? 0 : num_args - 4);
+	uint32_t entry_code[] = {
+		htole32(0xe58fe040), /*    0:    str      lr, [pc, #64]         */
+		htole32(0xe58fd040), /*    4:    str      sp, [pc, #64]         */
+		htole32(0xe59fd040), /*    8:    ldr      sp, [pc, #64]         */
+		htole32(0xe28fc040), /*    c:    add      ip, pc, #64           */
+		htole32(0xe1a0200d), /*   10:    mov      r2, sp                */
+		htole32(0xe49c0004), /*   14:    ldr      r0, [ip], #4          */
+		htole32(0xe3500000), /*   18:    cmp      r0, #0                */
+		htole32(0x0a000003), /*   1c:    beq      30 <entry+0x30>       */
+		htole32(0xe49c1004), /*   20:    ldr      r1, [ip], #4          */
+		htole32(0xe4821004), /*   24:    str      r1, [r2], #4          */
+		htole32(0xe2500001), /*   28:    subs     r0, r0, #1            */
+		htole32(0x1afffffb), /*   2c:    bne      20 <entry+0x20>       */
+		htole32(0xe8bc000f), /*   30:    ldm      ip!, {r0, r1, r2, r3} */
+		htole32(0xe12fff3c), /*   34:    blx      ip                    */
+		htole32(0xe59fe008), /*   38:    ldr      lr, [pc, #8]          */
+		htole32(0xe59fd008), /*   3c:    ldr      sp, [pc, #8]          */
+		htole32(0xe58f0000), /*   40:    str      r0, [pc]              */
+		htole32(0xe12fff1e), /*   44:    bx       lr                    */
+		htole32(0x00000000), /*   48:    .word    0x00000000            */
+		htole32(0x00000000), /*   4c:    .word    0x00000000            */
+	};
+
+	if (!soc_info)
+		return false;
+
+	/* Calculate the stack location */
+	new_sp = soc_info->scratch_addr +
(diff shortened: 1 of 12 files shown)


fel.c
```diff
@@ -512,9 +512,9 @@ uint32_t *aw_backup_and_disable_mmu(feldev_handle *dev,
 	 * checks needs to be relaxed).
 	 */
 
-	/* Basically, ignore M/Z/I/V/UNK bits and expect no TEX remap */
+	/* Ignore M/A/UNK/SW/Z/I/V/RR bits and expect no TEX remap */
 	sctlr = aw_get_sctlr(dev, soc_info);
-	if ((sctlr & ~((0x7 << 11) | (1 << 6) | 1)) != 0x00C50038)
+	if ((sctlr & ~((0x1f << 10) | (1 << 6) | (3 << 0))) != 0x00C50038)
```
ssvb: > fel: relax SCTLR check

> When entering FEL mode from later stages of U-Boot (via the go command,
> for instance), the SCTLR might have more bits set. This makes sunxi-fel
> to complain

And this is the expected behaviour. You can read more about it in #111 

> even though FEL mode works fine in this setup.

It only _seemingly_ works fine and this makes a huge difference. I don't want to have a long debate again and repeat what has been already said, but starting a bootloader from an undefined state is a bad practice that is universally frowned upon. Even Allwinner is doing a proper job in its switch to FEL code as mentioned earlier by @wens.

If you want to have a practical example, then a separate PMIC chip configures the CPU core voltage and retains its state after the CPU reset. This use case is already giving us some headaches because the initial state of the PMIC needs to be accounted for. And now we have a complex SoC with lots of peripherals, which can be configured in a lot of different ways.

Today somebody reports that entering FEL mode from later stages of U-Boot via the go command works "fine". But tomorrow people may report a bunch of obscure bugs while abusing this "feature". So a strong NAK from me.

> As the BROM FEL code has no issues with the Alignment bit set, allow
> bit 1 to be set as well. Allow bits 10 and 14 to be set as well, as they
> don't affect BROM code execution.

We have no need to care about these bits yet because they are not set in any of the supported configurations. This safety check may be triggered in two cases:
 * somebody is trying to add support for a new SoC
 * somebody is doing something wrong, just like in #111

The existence of #111 is a clear demonstration why this safety check is useful. If you want to improve the error message text for better clarity, then a patch would be welcome.

fel.c
```diff
@@ -947,8 +947,13 @@ void aw_fel_process_spl_and_uboot(feldev_handle *dev, const char *filename)
...
 #define SPL_SIGNATURE			"SPL" /* marks "sunxi" header */
-#define SPL_MIN_VERSION			1 /* minimum required version */
-#define SPL_MAX_VERSION			2 /* maximum supported version */
+#define SPL_MAJOR_BITS			3
+#define SPL_MINOR_BITS			5
+#define SPL_VERSION(maj, min)		\
+	((((maj) & ((1U << SPL_MAJOR_BITS) - 1)) << SPL_MINOR_BITS) | \
+	((min) & ((1U << SPL_MINOR_BITS) - 1)))
+#define SPL_MIN_VERSION			SPL_VERSION(0, 1)
+#define SPL_MAX_VERSION			SPL_VERSION(0, 31)
```
ssvb: > FEL: introduce semantic versioning for SPL header 

This patch is technically a bugfix for #104, but dressed as a grand new versioning scheme.

> Every addition of a new feature to the SPL header currently requires us
> to update the FEL tool, to teach it about the new supported maximum
> value.

Here you are implying that there was a policy, which required patching sunxi-tools on every SPL header version increase. I think that you actually need a citation to support this claim. It could be a message in a mailing list archive, some documentation in the sunxi-tools sources or in the wiki or at least a commit message,

In reality we need a _single_ simple bugfix to address the problem. Similar to what your patch is attempting to do. One could easily figure it out by doing a google search for the old discussions on this topic, filing a bug (kudos to the reporter of #104 who did it right) or asking a question.

When FEL-booting something via sunxi-tools, the following situations may potentially arise:
1. Something that is even not an U-Boot bootloader (maybe boot0 or one of the bare metal applications, such as uart0-helloworld-sdboot).
2. An old version of U-Boot, which does not support one of the required features (such as passing the `boot.scr` blob).
3. An older, but perfectly compatible version of the SPL header.
4. An equally matched version of the SPL header.
5. A newer, but still compatible version of the SPL header.
6. An new incompatible version of the SPL header.

Ideally, each of these situations needs to be handled gracefully. Providing comprehensive error messages in the case of failures, plus informative notifications in the verbose mode.

The old sunxi-tools code failed on the 5th case, because it incorrectly handled it as an error rather than just a helpful notification stating that _"The U-Boot image has a newer version of the SPL header, which means that there might be also a newer version of sunxi-tools available somewhere which supports more features"_ (intended to encourage packaging of the up to date versions of sunxi-tools by distro maintainers).

The old sunxi-tools code also made no distinction between the 1st and the 6th cases (a compatibility break in future versions of U-Boot could be done by changing the SPL_SIGNATURE). And this is where semantic versioning is supposed to provide a nice improvement, but it does not look like your patch takes a full advantage of it.

fel.c
apritzel: I don't agree, because I believe that a hacker's tool like sunxi-fel should not limit its users. And for that patch I spent half an hour over the ARM ARM to convince myself those SCTLR bits are safe. *BUT* I don't care enough to argue about that, so I will just drop that patch. I just wanted to be nice and close issue #111. @ssvb, can you please add a comment there and the close it as WONTFIX?

fel.c
Andre-ARM: @ssvb, I don't understand why you keep *talking* about this topic, when the problem is around for years now? Can't we just focus on a solution instead? I found it about time to just post a patch. If you don't think this is sufficient, fair enough, I am happy to adjust it to suit your needs or we take another solution instead. But I'd rather speak C instead of Markdown now. For the sake of getting those fixes in, I can drop this patch from this PR, so this can be merged.

But I would love to see a solution for this in a new release, especially as the next change is already lurking around the corner: https://lists.denx.de/pipermail/u-boot/2018-May/328805.html

apritzel:
Hi @ssvb, now that I dropped the "controversial" commits, can you please approve the remaining PR?
And sorry for not realising that "push --force"-ing the branch to let Github update the pull request apparently also kills the comment history :-(

Andre-ARM:
Ping, are there still outstanding issues with those remaining patches? @ssvb, could you please approve them, so that @paulkocialkowski could merge them?
Thanks! Andre

apritzel:
@wens, @rellla, @paulkocialkowski: can one of you please merge this PR (or comment on it)?
This gathers patches lying around for a while:
- SPI flash support, authored by @ssvb 
- H6 support, authored by @Icenowy
- minor fixes for various Github issues

This does not include any of the patches @ssvb had concerns about. People keep asking for a new release and the new features, so I'd like to make progress on this.
I am using both H6 and SPI flash support for many months now, I think others tested this as well.

Thanks,
Andre.

wens:
Merged.

PR #118, opened 2018-12-02
FEL: introduce semantic versioning for SPL header

Icenowy:
Every addition of a new feature to the SPL header currently requires us
to update the FEL tool, to teach it about the new supported maximum
value. Many times the FEL tool doesn't really care, but complains
anyway - and refuses to load.
Let's introduce semantic versioning [1] for this field, where backwards
compatible additions just increase a minor number, but incompatible
changes require bumping the major version.
We have 8 bits for the SPL header version, let's split this to have 3 bits
for the major and 5 bit for the minor version number.

[1] https://semver.org

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>

=========================

As the semantic versioning of SPL header is already accepted by
mainline U-Boot 2018.11, and the header version is raised to v0.3,
it's now needed to merge this patch into sunxi-tools, otherwise
sunxi-tools will fail to FEL boot U-Boot 2018.11.

Thanks
Icenowy

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -947,8 +947,13 @@ void aw_fel_process_spl_and_uboot(feldev_handle *dev, const char *filename)
  * the result is "true".
  */
 #define SPL_SIGNATURE			"SPL" /* marks "sunxi" header */
-#define SPL_MIN_VERSION			1 /* minimum required version */
-#define SPL_MAX_VERSION			2 /* maximum supported version */
+#define SPL_MAJOR_BITS			3
+#define SPL_MINOR_BITS			5
+#define SPL_VERSION(maj, min)		\
+	((((maj) & ((1U << SPL_MAJOR_BITS) - 1)) << SPL_MINOR_BITS) | \
+	((min) & ((1U << SPL_MINOR_BITS) - 1)))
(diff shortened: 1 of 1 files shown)


apritzel:
While I totally support this PR, I wonder if anyone sees any issues with the current code? The U-Boot build process keeps the version number at 2, it's only the SPL that increases it to 3 dynamically at runtime, in SRAM only. sunxi-fel shouldn't see it there, it just cares about the field in the SPL file header.
Of course we should merge it anyways, to prepare for the next version bump, which might be in the file header then.

EliasKotlyar:
I had to use this development version, because it was not supporting the newer u-boot version...

Icenowy:
@apritzel sunxi-fel will read back from SRAM to check the header version, regardless of the header version of the SPL image on the disk (and passed onto the device). So if SPL changes it to 3 in SRAM, sunxi-fel will read back 3.

This is to support the situation that write command is not executed at the same time with spl command.

Icenowy:
By the way, FEL booting of mainline U-Boot have been broken for at least 2 released versions.

I'm going to merge this pull request.

EliasKotlyar:
thanks for merging!

PR #122, opened 2019-03-03
meminfo: Access to io memory via pointers

vkochan:
The main reason for this is to be able compile with musl library,
because there is no support of inx/outx functions for ARM platform.

Signed-off-by: Vadim Kochan <vadim4j@gmail.com>

### DIFF
--- a/meminfo.c
+++ b/meminfo.c
@@ -22,7 +22,6 @@
 #include <sys/mman.h>
 #include <stdint.h>
 #include <errno.h>
-#include <sys/io.h>
 #include <stdbool.h>
 
 #include "common.h"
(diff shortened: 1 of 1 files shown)


meminfo.c
```diff
@@ -74,24 +73,26 @@ static enum sunxi_soc_version soc_version;
 unsigned int
 sunxi_io_read(void *base, int offset)
 {
-	return inl((unsigned long) (base + offset));
+	unsigned long port = (unsigned long) (base + offset);
```
apritzel: Minor nit: typically there is an empty line between variable declarations and code.

meminfo.c
```diff
@@ -74,24 +73,26 @@ static enum sunxi_soc_version soc_version;
...
+	return *((volatile unsigned long *) port);
```
apritzel: Can you either use `volatile unsigned int` or even better `volatile uint32_t`? The size of unsigned long depends on the C compiler, and here you want to explicitly do a 32-bit memory access.
For a start this otherwise would break on arm64 (admittedly this tool wouldn't work on those SoCs anyway, but still ...)

meminfo.c
```diff
@@ -74,24 +73,26 @@ static enum sunxi_soc_version soc_version;
...
 }
 
 void
 sunxi_io_write(void *base, int offset, unsigned int value)
 {
-	outl(value, (unsigned long) (base + offset));
+	unsigned long port = (unsigned long) (base + offset);
+	*((volatile unsigned long *) port) = value;
```
apritzel: Same two comments (empty line and uint32_t) as above

apritzel:
This seems to have been superseded by #144, which fixes the issues I pointed out above.

PR #123, opened 2019-04-08
spi: add support for H6

gamelaster:
Tested on Pine H64 rev B with Winbond W25Q128FW.

Since memory map on H6 has been heavily changed, new addresses has been required to add.
Actually, I'm not sure if this is the best implementation, it doesn't looks much good for me, until now it has been just sun4i and sun6i, but now there is also sun50i_h6, which using same approach as sun6i, but different addresses, so maybe to future it will be good to make struct with addresses?

Looking forward for reviews :)

### DIFF
--- a/fel-spiflash.c
+++ b/fel-spiflash.c
@@ -75,6 +75,10 @@ void fel_writel(feldev_handle *dev, uint32_t addr, uint32_t val);
 #define CCM_AHB_GATE_SPI0           (1 << 20)
 #define SUN6I_BUS_SOFT_RST_REG0     (0x01C20000 + 0x2C0)
 #define SUN6I_SPI0_RST              (1 << 20)
+#define SUN50I_H6_SPI_GATING_REG    (0x03001000 + 0x96C)
+#define SUN50I_H6_SPI0_RST          (1 << 16)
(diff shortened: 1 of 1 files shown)


gamelaster:
Bump

apritzel:
Hi Marek,

sorry, I didn't see your PR until after posting my version of that (#129) :-(
I agree with your own assessment ;-) that treating the H6 SPI as a separate implementation is not the best solution. My solution to that problem was to introduce a spi0_base field in soc_info, see https://github.com/linux-sunxi/sunxi-tools/pull/129/commits/d6866a8e54dbd77b014db865bd30fe9185150b5e.

Not sure how we want to take this from here, since you posted first, you deserve this being your contribution. Do you want to take my spi0_base patch, and rework your patch on top of that?
Or you could look and test my PR, if that is easier.

Cheers,
Andre.

gamelaster:
Hi @apritzel ,
it's okay, your PR is good in long-term, since is many changes in latest SoCs, but the register ofsets are almost same, so your solution is better. I knew that my implementation isn't perfect, since I made the v2 patch also for u-boot SPL SPI driver because of an issue you mentioning. I'm closing it in favor of #129 . Thanks :)

PR #129, opened 2020-01-09
H6 and R40 support for SPI flash access

apritzel:
The spiflash commands in sunxi-tools were not supporting the Allwinner H6 SoC so far, this series fixes this. While at it, add support for the R40 as well, since I have a board lying around here.
Tested on Pine H64 and Bananapi M2 Berry, regression tested on Pine64-LTS, OrangePi PC2 and OrangePi Zero.

Please review and test!

### DIFF
--- a/fel-spiflash.c
+++ b/fel-spiflash.c
@@ -173,6 +173,7 @@ static bool spi0_init(feldev_handle *dev)
 		gpio_set_cfgpin(dev, PC, 3, SUNXI_GPC_SPI0);
 		break;
 	case 0x1651: /* Allwinner A20 */
+	case 0x1701: /* Allwinner R40 */
 		gpio_set_cfgpin(dev, PC, 0, SUNXI_GPC_SPI0);
 		gpio_set_cfgpin(dev, PC, 1, SUNXI_GPC_SPI0);
 		gpio_set_cfgpin(dev, PC, 2, SUNXI_GPC_SPI0);
--- a/fel-spiflash.c
+++ b/fel-spiflash.c
@@ -85,7 +85,7 @@ void fel_writel(feldev_handle *dev, uint32_t addr, uint32_t val);
 #define SUN4I_CTL_RF_RST            (1 << 9)
 #define SUN4I_CTL_XCH               (1 << 10)
 
-#define SUN6I_TCR_XCH               (1 << 31)
+#define SUN6I_TCR_XCH               (1U << 31)
 
 #define SUN4I_SPI0_CCTL             (0x01C05000 + 0x1C)
 #define SUN4I_SPI0_CTL              (0x01C05000 + 0x08)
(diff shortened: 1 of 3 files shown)


fel-spiflash.c
```diff
@@ -153,13 +153,17 @@ static bool spi_is_sun6i(feldev_handle *dev)
...
+	uint32_t base;
 	soc_info_t *soc_info = dev->soc_info;
+
 	if (!soc_info) {
 		printf("Unable to fetch device information. "
 		       "Possibly unknown device.\n");
 		return false;
 	}
 
+	base = soc_info->spi0_base;
```
wens: Maybe add a check and fail if base == 0?

fel-spiflash.c
```diff
@@ -187,33 +192,45 @@ static bool spi0_init(feldev_handle *dev)
...
 				      SUN6I_SPI0_CCTL : SUN4I_SPI0_CCTL));
 	/* Choose 24MHz from OSC24M and enable clock */
-	writel((1U << 31), CCM_SPI0_CLK);
+	writel((1U << 31), is_h6(dev) ? CCM_H6_SPI0_CLK : CCM_SPI0_CLK);
 
 	if (spi_is_sun6i(dev)) {
 		/* Enable SPI in the master mode and do a soft reset */
 		reg_val = readl(base + SUN6I_SPI0_GCR);
-		reg_val |= (1 << 31) | 3;
+		reg_val |= (1U << 31) | 0x03;
```
wens: This is unrelated.

fel-spiflash.c
apritzel: Yeah, good idea.

fel-spiflash.c
apritzel: Right, my OCD kicking in whenever I see "1 << 31" (which I think is actually undefined behaviour). Might be worth fixing all of them in a separate patch.

apritzel:
There seems to be a sunxi specific, but unrelated problem in the Travis CI that makes those checks fail. Can someone please have a look? Seems like a patch to something is already applied.

plaes:
> There seems to be a sunxi specific, but unrelated problem in the Travis CI that makes those checks fail. Can someone please have a look? Seems like a patch to something is already applied.

Yeah, this failure is unrelated and consists of actually two issues:
1. We previously patched one of the fex files locally in this repository and this change has now been applied in the sunxi-boards repo
2. We currently don't support '-' and '/' characters in the fex files that have been recently added to sunxi-boards repo.

apritzel:
> 1. We previously patched one of the fex files locally in this repository and this change has now been applied in the sunxi-boards repo

Ah, thanks for the confirmation. Are you able to fix that?

> 2. We currently don't support '-' and '/' characters in the fex files that have been recently added to sunxi-boards repo.

There is PR #103 apparently addressing that. If that helps to unblock the problem, I can have a look at this (although I tried to ignore the existence of .fex altogether so far).

apritzel:
Abandoning this, as Icenowy got something something merged. Will push H6 and R40 support in a separate PR.

PR #131, opened 2020-04-19
Provide a wrapper for reset via watchdog

karlp:
The watchdog register isn't in the same place, nor uses the same values
to trigger a reset.

Signed-off-by: Karl Palsson <karlp@tweak.net.au>

Incomplete as far as filling in the structure for all the socs, if there's support for this, I can go and check more datasheets, but I didn't want to go and do all that datasheet trawling up front.  I only have a h3 board locally to test with unfortunately.  Inspiration from https://linux-sunxi.org/FEL#Tips_and_tricks

### DIFF
--- a/soc_info.c
+++ b/soc_info.c
@@ -110,6 +110,16 @@ sram_swap_buffers h6_sram_swap_buffers[] = {
 	{ .size = 0 }  /* End of the table */
 };
 
+const watchdog_info wd_a10_compat = {
+	.reg_mode = 0x01C20C94,
+	.reg_mode_value = 3,
+};
+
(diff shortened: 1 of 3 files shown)


soc_info.c
```diff
@@ -110,6 +110,16 @@ sram_swap_buffers h6_sram_swap_buffers[] = {
 	{ .size = 0 }  /* End of the table */
 };
 
+watchdog_info wd_a10_compat = {
```
plaes, line 113: const?

karlp: sure, seems reasonable :)

PR #144, opened 2020-10-08
meminfo: Replace sys/io.h by direct register accesses.

daym:
Signed-off-by: Danny Milosavljevic <dannym@scratchpost.org>

sys/io.h has been removed in glibc 2.30.

https://sourceware.org/legacy-ml/libc-alpha/2019-08/msg00029.html

* On 32-bit Arm, support for the port-based I/O emulation and the <sys/io.h>
  header have been removed.

It's easy to just directly access--so I did that...

### DIFF
--- a/meminfo.c
+++ b/meminfo.c
@@ -22,7 +22,6 @@
 #include <sys/mman.h>
 #include <stdint.h>
 #include <errno.h>
-#include <sys/io.h>
 #include <stdbool.h>
 
 #include "common.h"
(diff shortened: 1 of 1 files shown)


apritzel (approved):
So @vkochan had a similar patch #122, but he did not reply to the comments, and this version here seems to address them all fine.
So we should go ahead with this one. Just compile tested (and I had issues already on Ubuntu 18.04's glibc 2.27).

PR #149, opened 2020-11-08
wdreset: Add remaining SoCs

apritzel:
The "wdreset" command so far only covered a few SoCs.

Add the watchdog data for the other ones as well.

### DIFF
--- a/soc_info.c
+++ b/soc_info.c
@@ -141,6 +141,16 @@ const watchdog_info wd_h3_compat = {
 	.reg_mode_value = 1,
 };
 
+const watchdog_info wd_a80 = {
+	.reg_mode = 0x06000CB8,
+	.reg_mode_value = 1,
+};
+
+const watchdog_info wd_h6_compat = {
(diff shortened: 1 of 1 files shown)


apritzel:
@plaes @karlp Can you please have a look at this one? The info was taken from an old patch of mine (never posted): https://github.com/apritzel/sunxi-tools/commit/4c90e1a6a492ab

karlp:
Looks legit, but I've got no hw to test against, and haven't read any of the datasheets for the other parts.

apritzel:
@karlp Many thanks for having a look. I just tested this on A64, H6, H2+ and R40 right now, and others (H5) before.
You can compare U-Boot's reset_cpu() implementation, which does a similar differentiation between GEN_SUN4I and GEN_SUN6I, that should match to wd_a10_compat and wd_h3_compat, with some exceptions:
https://github.com/u-boot/u-boot/blob/master/arch/arm/mach-sunxi/board.c#L318-L347

PR #150, opened 2020-11-10
Enabling SPI flash support for H6 & R40 (+ fixes)

apritzel:
H6 and R40 support are fairly trivial to add (now). Also two fixes I came up with during development.

### DIFF
--- a/fel-spiflash.c
+++ b/fel-spiflash.c
@@ -245,14 +245,7 @@ static bool spi0_init(feldev_handle *dev)
 		reg_val = readl(H6_CCM_SPI_BGR);
 		reg_val |= H6_CCM_SPI0_GATE_RESET;
 		writel(reg_val, H6_CCM_SPI_BGR);
-
-		/* 24MHz from OSC24M */
-		writel((1 << 31), H6_CCM_SPI0_CLK);
 	} else {
-		reg_val = readl(CCM_AHB_GATING0);
-		reg_val |= CCM_AHB_GATE_SPI0;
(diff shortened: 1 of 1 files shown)


fel-spiflash.c
```diff
@@ -223,10 +226,13 @@ static bool spi0_init(feldev_handle *dev)
...
 		break;
 	case 0x1817: /* Allwinner V831 */
+		gpio_set_cfgpin(dev, PC, 1, SUN50I_GPC_SPI0);	/* SPI0-CS */
+		/* fall-through */
+	case 0x1728: /* Allwinner H6 */
 		gpio_set_cfgpin(dev, PC, 0, SUN50I_GPC_SPI0);
-		gpio_set_cfgpin(dev, PC, 1, SUN50I_GPC_SPI0);
 		gpio_set_cfgpin(dev, PC, 2, SUN50I_GPC_SPI0);
 		gpio_set_cfgpin(dev, PC, 3, SUN50I_GPC_SPI0);
+		gpio_set_cfgpin(dev, PC, 5, SUN50I_GPC_SPI0);	/* H6: CS */
```
wens, line 235: What's PC5 on V831? I'd like to avoid muxing unused pins if possible.

wens (commented):
The rest looks good.

fel-spiflash.c
apritzel: PC5/010 is SPI0-HOLD. So configuring this is harmless, it just means it's under the control of the SPI controller. I can extend the comment.

PR #151, opened 2020-11-19
spi: Add H616 support

apritzel:
The H616 SPI is very similar to the H6, only differs in the GPIOs
(again).
Add the SoC-ID at the right places and add the GPIOs according to the
manual.

Tested on OrangePi Zero 2.


Icenowy (approved):
LGTM.

PR #152, opened 2020-12-23
sunxi-fel FIT image support

apritzel:
This adds support for the "uboot" command to deal with a FIT image, as used on the 64-bit U-Boot builds. Together with the preliminary [64-bit U-Boot FEL support](https://lists.denx.de/pipermail/u-boot/2020-November/433083.html) this allows to bootstrap 64-bit boards with a simple:
`sunxi-fel uboot u-boot-sunxi-with-spl.bin`
This adds a build dependency to libfdt for sunxi-fel. We could make this optional, if this becomes a problem. I am open to suggestions on how to best autodetect this in the Makefile (there are some try-build macros in other projects we can copy from).

### DIFF
--- a/fit_image.c
+++ b/fit_image.c
@@ -1,0 +1,282 @@
+/*
+ * Copyright (C) 2018-2020  Andre Przywara <osp@andrep.de>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; under version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdio.h>
+#include <stdint.h>
+#include <libfdt.h>
+
+#include "common.h"
+#include "fel_lib.h"
+#include "fit_image.h"
+
+/* defined in fel.c */
+extern bool verbose;
+
+#define IH_ARCH_INVALID			0
+#define IH_ARCH_ARM			2
+#define IH_ARCH_ARM64			22
+
+#define IH_OS_INVALID			0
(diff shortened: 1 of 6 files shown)


fit_image.c
```diff
@@ -0,0 +1,288 @@
...
+	     prop && (str - prop->data) < len && *str;
+	     str += strlen(str) + 1) {
+		uint32_t addr;
+
+		if (fit_get_image_info(fit, str, &img)) {
+			printf("Can't load loadable \"%s\", skipping.\n", str);
+			continue;
+		}
+		addr = fit_load_image(dev, &img);
+		if (addr == ~0U)
```
smaeul, line 261: And here as well

fit_image.c
```diff
@@ -0,0 +1,288 @@
...
+
+	entry_arch = IH_ARCH_INVALID;
+	dtb_addr = 0;
+
+	/* Load the image described as "firmware". */
+	str = fdt_getprop_str(fit, node, "firmware");
+	if (str && !fit_get_image_info(fit, str, &img)) {
+		uint32_t addr = fit_load_image(dev, &img);
+
+		if (addr == ~0U)
```
smaeul, line 240: This condition appears to be unreachable

plaes:
Tested with Olimex A64 Olinuxino, therefore:
`Tested-by: Priit Laes <plaes@plaes.org>`

> This adds a build dependency to libfdt for sunxi-fel. We could make this optional, if this becomes a problem. I am open to suggestions on how to best autodetect this in the Makefile (there are some try-build macros in other projects we can copy from).

I don't think this would be an issue as this tool is mostly targeting developers. And libfdt does not provide a pkgconf(ig) files anyway (at least on Debian and Ubuntu).

jernejsk:
Tested-by: Jernej Skrabec <jernej.skrabec@siol.net>

fit_image.c
apritzel: Yeah, a futile attempt to teach this tool some error handling aside from "panic". But even fel_**lib**.c doesn't justify its name, so I will remove this.

apritzel:
Hi, is there anything I can do to move this forward? There are two Tested-by's above, and I addressed @smaeul's comments.
I think there is little danger of regression, as the original U-Boot legacy image handling is untouched by this series.
So shall we give this a go, and open up the testing base for the FIT image functionality by merging it?

PR #154, opened 2020-12-31
fel: Allow larger SPL payload

apritzel:
At the moment we artificially restrict the payload of the "sunxi-fel spl" command to 32KB, as this was the hardcoded limit of older BootROMs. However newer SoCs (starting with the H6) don't have this limit anymore, and FEL was actually never bound to this anyway, as we can load larger sizes easily.
The H616 needs (and supports) a bigger SPL, so we need to lift this artificial limitation.

This PR introduces a "sram_size" variable, which holds the maximum usable amount of contiguous SRAM at the SPL load address. We initialise our existing spl_len_limit with this value, then let the existing code restrict this further, by observing the location of the backup and thunk buffers.

This alone already increases the maximum SPL size for some SoCs, but by choosing the backup buffer locations more cleverly, we can increase the maximum SPL size even further.
This is done for the H616, H6, A64 and H5, as I could test them easily.
Other SoCs can have the size increased as well, typically by experimentation (as the BROM might use some of the existing SRAM for its own purposes).


jernejsk:
Works fine with H616, which requires bigger SPL size.

Tested-by: Jernej Skrabec <jernej.skrabec@siol.net>

PR #159, opened 2021-07-15
R329 support

Icenowy:
This pull request contains support for Allwinner R329 chip.

The first patch is for supporting FEL SPL, and the second is for porting uart0-helloworld-sdboot.


apritzel (approved):
Hi, looks alright on the first glance. I don't have an R329 board, but don't see a problem with merging that. Briefly regression tested on an H5 board.

PR #164, opened 2021-11-28
Add V536 SoC support

danielkucera:


fel-spiflash.c
```diff
@@ -228,6 +231,8 @@ static bool spi0_init(feldev_handle *dev)
 		gpio_set_cfgpin(dev, PC, 2, SUN50I_GPC_SPI0);
 		gpio_set_cfgpin(dev, PC, 3, SUN50I_GPC_SPI0);
 		break;
+	case 0x1816: /* Allwinner V536 */
+		gpio_set_cfgpin(dev, PC, 1, SUN50I_GPC_SPI0);	/* SPI0-CS */
```
apritzel, line 235: there is no break or fallthrough comment here. It looks like the V536 is the same as the V831 when it comes to SPI, so just adding the "case 0x1816:" line should be enough? See the H6 and H616 above for an example.

soc_info.c
```diff
@@ -127,6 +127,11 @@ sram_swap_buffers v831_sram_swap_buffers[] = {
 	{ .size = 0 }  /* End of the table */
 };
 
+sram_swap_buffers v536_sram_swap_buffers[] = {
```
apritzel, line 130: This looks identical to the V831 structure above, so no need to repeat this here. Just reference v831_sram_swap_buffers below instead.

soc_info.c
```diff
@@ -331,6 +336,16 @@ soc_info_t soc_info_table[] = {
...
 	},{
+		.soc_id       = 0x1816, /* Allwinner V536 */
+		.name         = "V536",
+		.spl_addr     = 0x20000,
+		.scratch_addr = 0x21000,
+		.thunk_addr   = 0x2A200, .thunk_size = 0x200,
+		.swap_buffers = v536_sram_swap_buffers,
+		.sid_base     = 0x03006000,
+		.sid_offset   = 0x200,
+		.rvbar_reg    = 0x09010040,
```
apritzel, line 347: rvbar_reg is only for 64-bit SoCs.
You can probably copy the whole V831 definition here, just change .soc_id and the name. That should add .sram_size and .watchdog, you can verify the latter with "sunxi-fel wdreset".

apritzel (commented):
Hi, thanks for the patch!
Can you please change the subject to speak of general V536 support, because it's more than the SPI flash support. And mention there that this SoC is very close to the V831, probably identical for the purpose of sunxi-fel.
Please note the individual comments.

Cheers,
Andre

fel-spiflash.c
danielkucera: correct. fixed.

soc_info.c
danielkucera: fixed.

soc_info.c
danielkucera: I meant to copy 0x1817 but accidentally copied 0x1823. Fixed in latest commit. wdreset verified to work. thanks.

apritzel (approved):
Can you please still change the subject line to mention general V536 support, and mention in the commit log that this is basically the same as the V831?

danielkucera:
I need to mention that this was actually tested on V316 but the chip has identical soc_id so I presume there are only changes to video HW blocks (also according to product brief)

danielkucera:
Ping.

apritzel (approved):
Looks good, thanks for the changes.

akalime:
Submitting hardware-verified PR for V536 (0x1816). Tested on real V536 
silicon (Eonon T86PRO_V101). Key finding: thunk_addr is 0x2E000, not 
0x2A200 as assumed from V831 in the 2021 attempt.

PR #170, opened 2022-01-29
Add F1C100s support

giuliobenetti:
I've forked @Icenowy branch, rebased and adapted to make ARMv7+ and ARMv5TE to cohexist.

### DIFF
--- a/fel-spiflash.c
+++ b/fel-spiflash.c
@@ -85,6 +85,7 @@ void fel_writel(feldev_handle *dev, uint32_t addr, uint32_t val);
 #define H6_CCM_SPI_BGR              (0x03001000 + 0x96C)
 #define H6_CCM_SPI0_GATE_RESET      (1 << 0 | 1 << 16)
 
+#define SUNIV_GPC_SPI0              (2)
 #define SUNXI_GPC_SPI0              (3)
 #define SUN50I_GPC_SPI0             (4)
 
@@ -96,27 +97,30 @@ void fel_writel(feldev_handle *dev, uint32_t addr, uint32_t val);
 
 #define SUN6I_TCR_XCH               (1U << 31)
 
-#define SUN4I_SPI0_CCTL             (spi_base(dev) + 0x1C)
-#define SUN4I_SPI0_CTL              (spi_base(dev) + 0x08)
-#define SUN4I_SPI0_RX               (spi_base(dev) + 0x00)
-#define SUN4I_SPI0_TX               (spi_base(dev) + 0x04)
-#define SUN4I_SPI0_FIFO_STA         (spi_base(dev) + 0x28)
-#define SUN4I_SPI0_BC               (spi_base(dev) + 0x20)
-#define SUN4I_SPI0_TC               (spi_base(dev) + 0x24)
+static uint32_t spi0_base;
 
-#define SUN6I_SPI0_CCTL             (spi_base(dev) + 0x24)
-#define SUN6I_SPI0_GCR              (spi_base(dev) + 0x04)
-#define SUN6I_SPI0_TCR              (spi_base(dev) + 0x08)
-#define SUN6I_SPI0_FIFO_STA         (spi_base(dev) + 0x1C)
-#define SUN6I_SPI0_MBC              (spi_base(dev) + 0x30)
-#define SUN6I_SPI0_MTC              (spi_base(dev) + 0x34)
-#define SUN6I_SPI0_BCC              (spi_base(dev) + 0x38)
-#define SUN6I_SPI0_TXD              (spi_base(dev) + 0x200)
-#define SUN6I_SPI0_RXD              (spi_base(dev) + 0x300)
+#define SUN4I_SPI0_CCTL             (spi0_base + 0x1C)
+#define SUN4I_SPI0_CTL              (spi0_base + 0x08)
+#define SUN4I_SPI0_RX               (spi0_base + 0x00)
+#define SUN4I_SPI0_TX               (spi0_base + 0x04)
+#define SUN4I_SPI0_FIFO_STA         (spi0_base + 0x28)
+#define SUN4I_SPI0_BC               (spi0_base + 0x20)
+#define SUN4I_SPI0_TC               (spi0_base + 0x24)
+
+#define SUN6I_SPI0_CCTL             (spi0_base + 0x24)
+#define SUN6I_SPI0_GCR              (spi0_base + 0x04)
+#define SUN6I_SPI0_TCR              (spi0_base + 0x08)
+#define SUN6I_SPI0_FIFO_STA         (spi0_base + 0x1C)
+#define SUN6I_SPI0_MBC              (spi0_base + 0x30)
+#define SUN6I_SPI0_MTC              (spi0_base + 0x34)
+#define SUN6I_SPI0_BCC              (spi0_base + 0x38)
+#define SUN6I_SPI0_TXD              (spi0_base + 0x200)
+#define SUN6I_SPI0_RXD              (spi0_base + 0x300)
 
 #define CCM_SPI0_CLK_DIV_BY_2       (0x1000)
 #define CCM_SPI0_CLK_DIV_BY_4       (0x1001)
 #define CCM_SPI0_CLK_DIV_BY_6       (0x1002)
+#define CCM_SPI0_CLK_DIV_BY_32      (0x100f)
 
 static uint32_t gpio_base(feldev_handle *dev)
 {
@@ -128,24 +132,6 @@ static uint32_t gpio_base(feldev_handle *dev)
 		return 0x0300B000;
 	default:
 		return 0x01C20800;
-	}
-}
-
-static uint32_t spi_base(feldev_handle *dev)
-{
-	soc_info_t *soc_info = dev->soc_info;
-	switch (soc_info->soc_id) {
-	case 0x1623: /* A10 */
-	case 0x1625: /* A13 */
-	case 0x1651: /* A20 */
-	case 0x1701: /* R40 */
-		return 0x01C05000;
-	case 0x1817: /* V831 */
-	case 0x1728: /* H6 */
-	case 0x1823: /* H616 */
-		return 0x05010000;
-	default:
-		return 0x01C68000;
 	}
 }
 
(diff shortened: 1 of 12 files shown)


Mr-Bossman:
Tested on f1c100s and 200s works fine!!!

apritzel:
Hi, I made simplified version of this here: https://github.com/apritzel/sunxi-tools/commits/f1c100
There is really no need (yet) to go through all the hassle of uploading different code for v7 and v5, if the v5 version works on the v7/v8 parts as well.
Reportedly this doesn't work, though (the reason why I didn't push it yet). It's not easy to debug without h/w, though, which is still underway.
Can one of you possibly look into this? Or adjust your series accordingly?

fel-spiflash.c
```diff
@@ -96,27 +97,30 @@ void fel_writel(feldev_handle *dev, uint32_t addr, uint32_t val);
...
+
+#define SUN6I_SPI0_CCTL             (spi0_base + 0x24)
+#define SUN6I_SPI0_GCR              (spi0_base + 0x04)
+#define SUN6I_SPI0_TCR              (spi0_base + 0x08)
+#define SUN6I_SPI0_FIFO_STA         (spi0_base + 0x1C)
+#define SUN6I_SPI0_MBC              (spi0_base + 0x30)
+#define SUN6I_SPI0_MTC              (spi0_base + 0x34)
+#define SUN6I_SPI0_BCC              (spi0_base + 0x38)
+#define SUN6I_SPI0_TXD              (spi0_base + 0x200)
+#define SUN6I_SPI0_RXD              (spi0_base + 0x300)
```
apritzel, line 118: That whole part looks like wouldn't be necessary? We gained support for different base addresses meanwhile, which is now handled in the spi_base() function. So you can just keep the old code here, and you should be good, as you also added the SoC ID to that function.
Or is there something I am missing?

giuliobenetti: @apritzel i will check it tomorrow.
I didn’t know your nickname :-)
I didn’t think that armv5te code is compatible to armv7+.
Yours is the simplest solution!
So I’ll check, eventually fixed if needed and open a new PR.

Thanks for pointing!

giuliobenetti:
> Tested on f1c100s and 200s works fine!!!

Thank you

fel-spiflash.c
apritzel: Sorry, but this is not how it works. This change of the registers is just a rebase gone wrong, it should be fixed now. You can just fix it, and do a git push --force, the rest is done by github (updating the PR).

giuliobenetti: @apritzel i know it’s like that. I was talking about the merging of assembly code. I do it now.

giuliobenetti: @apritzel I was on mobile, now I have checked well on pc.
We need those defines, they changed:
-#define SUN4I_SPI0_CCTL             (spi_base(dev) + 0x1C)
+#define SUN4I_SPI0_CCTL             (spi0_base + 0x1C)

spi_base() must be removed instead because it's not used by anyone.
I'm going to fix in a moment.

Thank you for reviewing!

giuliobenetti:
> Hi, I made simplified version of this here: https://github.com/apritzel/sunxi-tools/commits/f1c100 There is really no need (yet) to go through all the hassle of uploading different code for v7 and v5, if the v5 version works on the v7/v8 parts as well. Reportedly this doesn't work, though (the reason why I didn't push it yet). It's not easy to debug without h/w, though, which is still underway. Can one of you possibly look into this? Or adjust your series accordingly?

Pardon, I wanted to answer here by saying that your optimization can be applied later.
At the moment this PR works ans is tested, so maybe you or @Icenowy can add an optimization later.
Is it ok?

fel-spiflash.c
```diff
@@ -131,24 +135,6 @@ static uint32_t gpio_base(feldev_handle *dev)
 	}
 }
 
-static uint32_t spi_base(feldev_handle *dev)
```
giuliobenetti, line 134: @apritzel I had to remove these lines, not the #defines. Now it should be ok!
What do you think?

apritzel: Actually this got worse, I think you broke H6 and friends now :-)
I believe the solution is much simpler than you think: Just remove the large hunk at the beginning. If you look closely, that already consults a runtime chosen variable, and you add the F1C100s to the bunch with a single "case" line spi_base(). Mission accomplished, no?

fel-spiflash.c
```diff
@@ -267,12 +268,6 @@ static bool spi0_init(feldev_handle *dev)
 		writel(reg_val, CCM_AHB_GATING0);
 	}
 
-	/* divide by 4 */
-	writel(CCM_SPI0_CLK_DIV_BY_4, spi_is_sun6i(dev) ? SUN6I_SPI0_CCTL :
-							  SUN4I_SPI0_CCTL);
-	/* Choose 24MHz from OSC24M and enable clock */
-	writel(1U << 31, soc_is_h6_style(dev) ? H6_CCM_SPI0_CLK : CCM_SPI0_CLK);
-
```
apritzel, line 275: Why has this part moved? I think this reverts commit 205e208a70cf6. Another rebase artifact?

giuliobenetti: @apritzel damn, you're right! Too many things to do in parallel and then I do such mistakes.
Thank you for pointing me. I fix and git push --force.

apritzel:
I don't understand what is the rush, those patches linger around for ages, and I think we should take a few more days to get them right. It is not just about an "optimisation", this series introduces quite some elaborate changes that are most likely unnecessary. I can probably try to find the problem with my shorter version by "code starring", but I also need to sleep at times ;-)
And also the SPI flash support seems to break things now, and re-introduces issues. I think what you miss is most of the problems that the original patch addressed have been solved upstream meanwhile (for the H6 support), in a slightly different fashion, that makes it actually easier to add new chips. So the new support patch is much smaller and easier than what you suggest.

giuliobenetti:
I close this PR in favor of @apritzel branch

apritzel:
@giuliobenetti I am sorry, I didn't want to brush you off, but if there are issues, we need to discuss and fix them.
I updated my branch, and added SPI flash support. See the top patch for what I meant earlier about the SPI support being easier: https://github.com/apritzel/sunxi-tools/commits/f1c100
Can you briefly test whether this branch works for you? And even if not, if you cherry-pick the top three patches to your branch (removing the respective patches in your version), you should get the cleaner SPI flash support.

giuliobenetti:
Hi Andre,

On 01/02/22 02:51, Andre Przywara wrote:
> @giuliobenetti <https://github.com/giuliobenetti> I am sorry, I didn't 
> want to brush you off, but if there are issues, we need to discuss and 
> fix them.

Sure, don't worry, I was too in a hurry on this topic and I didn't focus 
there was a better and easier way to do it.
Also while rebasing(in a hurry again :-/) I've left back more than one 
mistake.

> I updated my branch, and added SPI flash support. See the top patch for 
> what I meant earlier about the SPI support being easier: 
> https://github.com/apritzel/sunxi-tools/commits/f1c100 
> <https://github.com/apritzel/sunxi-tools/commits/f1c100>

Yes, I've cloned it yesterday.

> Can you briefly test whether this branch works for you?  > And even if not,
> if you cherry-pick the top three patches to your branch (removing the 
> respective patches in your version), you should get the cleaner SPI 
> flash support.

Sure, I will go with that in few days. The goal is to make it work.
I've already pushed your branch to my repo:
https://github.com/giuliobenetti/sunxi-tools/tree/dev/port-apritzel-f1c100s

And so I will go on from there and, if needed, fix it.

Thank you!
-- 
Giulio Benetti
Benetti Engineering sas

apritzel:
@giuliobenetti just a heads up: I pushed a new version last night, so what you branched into your repo is outdated now. I stared at the code yesterday, but couldn't spot any obvious issues. It certainly still works on H3 and H6, so I believe the new thunk is good. gas didn't complain, so all instructions should be ARMv5 safe. There could be something with the F1C100s socinfo, but I don't see where.

giuliobenetti:
Hi @apritzel,

> @giuliobenetti just a heads up: I pushed a new version last night, so what you branched into your repo is outdated now. I stared at the code yesterday, but couldn't spot any obvious issues. It certainly still works on H3 and H6, so I believe the new thunk is good. gas didn't complain, so all instructions should be ARMv5 safe. There could be something with the F1C100s socinfo, but I don't see where.

I've just tested this branch:
https://github.com/apritzel/sunxi-tools/tree/f1c100
and it works perfectly with F1C100s LicheePi Nano.
I've tried spiflash-write/read/info while zeroing, writing random stuff at more offsets and read back and compare and it just works :)
It gets recognized with no problem, both F1C100s and F1C200s. Only on my 200s board I don't have a spiflash soldered, so I
can't test spiflash-* there but since soc_is the same and judgin from User Manuals, for me they are totally the same on spiflash side.
So for me this is a go fo you to open a PR with the branch above.
You've been great considering you had no boards to test on!

Best regards!

giuliobenetti:
@apritzel last thing I've noticed, and I think it's correct, is that the part of block not filled with anything, for example if I have a file of 100 bytes, is that on not written block I find 0xFF and for me it's correct since that is the block erase command that worked, I don't see any sense to fill those other bytes with zeroes.

apritzel:
@giuliobenetti As I have received a LicheePi Nano, I managed to debug and fix the SPL issue: the ISB instruction is not implemented in the ARM926 core of the SoC, so that hangs. Making this conditional for just ARMv7+ cores fixes that issue. But also the mainline U-Boot port had a bug, which prevented return to FEL. This is now fixed as well (just not merged yet).
I pushed a fixed version of my branch (https://github.com/apritzel/sunxi-tools/tree/f1c100), can you please test this? I need to do further regression testing on more boards, and will then send a PR.

giuliobenetti:
Hi @apritzel, unfortunately I've just rebuilt uboot[1] and your sunxi-tools[2] but I still have the same problem:
sudo ./sunxi-fel spl/uboot u-boot-sunxi-with-spl.bin
but then sunxi-fel still hangs anf fails with:
usb_bulk_send() ERROR -7: Operation timed out

[1]: https://source.denx.de/u-boot/custodians/u-boot-sunxi
[2]: https://github.com/apritzel/sunxi-tools/tree/f1c100

PR #172, opened 2022-02-20
Add license header to tests/*

ijc:
These files were all originally by Bernhard in 2016 via these commits:

8445d71 unify-fex: Use int64_t and portable format specifiers for its output
f957f89 fextest: Add a dedicated rule for (CI-)testing sunxi-boards
6ec3876 tests: Improve code coverage by testing corner cases
cd5a0a3 tests: Improve on testing fexc
5cbb0c9 tests: Introduce a basic testing framework

and have not otherwise been touched. The header was added to unify-fex.c when
it was added in cd5a0a3303cb ("tests: Improve on testing fexc") but not to the
other files.

The license in unify-fex.c is the GPLv2+ which corresponds to the content of
LICENSE.md both at the time and now. I think therefore that it is appropriate
to duplicate the header into the other files (changing only the character used
for comments to suit)

--- 

I think it's pretty clear with/without the explicit header what the intended is, but Debian likes me (as package maintainer) to be a bit pedantic about this sort of thing (see also: 227c7e031f24ee829d1eb81749ea3d75b35bb72f). 

/cc @n1tehawk

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,6 +1,20 @@
 #
 # tests/Makefile
 #
+# Copyright (C) 2016  Bernhard Nortmann <bernhard.nortmann@web.de>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


apritzel (approved):
Looks good to me, the license text is the same in every file and matches the one from the commit you mentioned.
I feel like we should go to SPDX, but that's something for another time.

PR #173, opened 2022-03-06
Allwinner F1C100 support

apritzel:
Hi, this is based on ideas by @Icenowy (https://github.com/Icenowy/sunxi-tools/commits/f1c100s-spiflash), but changes her multiple architectures approach to just making the existing code both ARMv7 and ARMv5 compatible, so we still get away with just one SPL thunk, plus building the other code for ARMv5.
On top of the SPL functionality, this enables SPI flash access on the F1C100, since this is currently the only way to get access to the flash, and SPI booting works already with mainline U-Boot (PR pending).
Supersedes PR #170. On top of my github-actions branch, for which there is PR #172 pending.
@giuliobenetti @Mr-Bossman can you please confirm that this works for you?


giuliobenetti:
Hi @apritzel, i've tested it with both F1C100s on LicheePI-Nano and F1C200s on @Mr-bossman board and sunxi-tools work
perfectly, including spl/uboot, spiflash-* commands etc.

Thank you for the effort!
Best regards

PR #178, opened 2022-07-05
V853 support

Icenowy:
This PR contains V853 support. (It comes even before V853 document release...)

V853 BROM enables I-Cache, but without FEL write command invalidating it, which makes uploading different thunks to the same address not working. It's currently worked around by disabling I-Cache.

In addition, V853 has a GPIO controller with a new register map (the same with D1), support for this GPIO controller is also added to uart0-helloworld-sdboot.

### DIFF
--- a/fel_lib.c
+++ b/fel_lib.c
@@ -40,6 +40,7 @@ struct _felusb_handle {
 	libusb_device_handle *handle;
 	int endpoint_out, endpoint_in;
 	bool iface_detached;
+	bool icache_hacked;
 };
 
 /* a helper function to report libusb errors */
@@ -214,7 +215,7 @@ void aw_fel_read(feldev_handle *dev, uint32_t offset, void *buf, size_t len)
 }
 
 /* AW_FEL_1_WRITE request */
-void aw_fel_write(feldev_handle *dev, const void *buf, uint32_t offset, size_t len)
+static void aw_fel_write_raw(feldev_handle *dev, const void *buf, uint32_t offset, size_t len)
 {
 	if (len == 0)
 		return;
@@ -229,6 +230,33 @@ void aw_fel_execute(feldev_handle *dev, uint32_t offset)
 {
 	aw_send_fel_request(dev, AW_FEL_1_EXEC, offset, 0);
 	aw_read_fel_status(dev);
+}
+
+static void aw_disable_icache(feldev_handle *dev)
+{
+	soc_info_t *soc_info = dev->soc_info;
+	uint32_t arm_code[] = {
+		/* Clear SCTLR.I */
+		htole32(0xee110f10), /* mrc 15, 0, r0, cr1, cr0, {0} ;SCTLR */
+		htole32(0xe3c00a01), /* bic r0, r0, #0x1000 */
+		htole32(0xee010f10), /* mcr 15, 0, r0, cr1, cr0, {0} ;SCTLR */
+		/* Invalidate I-Cache */
+		htole32(0xee070f15), /* mcr 15, 0, r0, cr7, cr5, {0} ;ICIALLU */
+		/* Barrier to force instruction refetching */
+		htole32(0xf57ff06f), /* isb sy */
+		htole32(0xe12fff1e), /* bx  lr */
+	};
+	aw_fel_write_raw(dev, arm_code, soc_info->scratch_addr, sizeof(arm_code));
+	aw_fel_execute(dev, soc_info->scratch_addr);
+}
+
+void aw_fel_write(feldev_handle *dev, const void *buf, uint32_t offset, size_t len)
+{
+	if (dev->soc_info->icache_fix && !dev->usb->icache_hacked) {
+		aw_disable_icache(dev);
+		dev->usb->icache_hacked = true;
+	}
+	aw_fel_write_raw(dev, buf, offset, len);
 }
 
 /*
--- a/soc_info.c
+++ b/soc_info.c
@@ -182,6 +182,11 @@ const watchdog_info wd_h6_compat = {
 	.reg_mode_value = 1,
 };
 
+const watchdog_info wd_v853_compat = {
+	.reg_mode = 0x020500b8,
+	.reg_mode_value = 0x16aa0001,
+};
+
 soc_info_t soc_info_table[] = {
 	{
 		.soc_id       = 0x1623, /* Allwinner A10 */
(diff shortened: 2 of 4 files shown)


fel_lib.c
```diff
@@ -231,6 +232,33 @@ void aw_fel_execute(feldev_handle *dev, uint32_t offset)
 	aw_read_fel_status(dev);
 }
 
+static void aw_disable_icache(feldev_handle *dev)
+{
+	soc_info_t *soc_info = dev->soc_info;
+	uint32_t arm_code[] = {
+		htole32(0xee110f10), /* mrc     15, 0, r0, cr1, cr0, {0} */
```
apritzel, line 239: Please add the register name in the comment (SCTLR)

fel_lib.c
```diff
@@ -231,6 +232,33 @@ void aw_fel_execute(feldev_handle *dev, uint32_t offset)
...
+		htole32(0xe3c00a01), /* bic     r0, r0, #0x1000          */
+		htole32(0xee010f10), /* mcr     15, 0, r0, cr1, cr0, {0} */
+		htole32(0xe3a00000), /* mov     r0, #0                   */
```
apritzel, line 242: this is not needed, the Rt content is ignored

fel_lib.c
```diff
@@ -231,6 +232,33 @@ void aw_fel_execute(feldev_handle *dev, uint32_t offset)
...
+		htole32(0xee070f15), /* mcr     15, 0, r0, cr7, cr5, {0} */
+		htole32(0xee070fd5), /* mcr     15, 0, r0, cr7, cr5, {6} */
```
apritzel, line 244: again please have the register names in the comment. And did you *need* the explicit BPIALL? According to the ARMv7 ARM:
"ICIALLU: Invalidate all instruction caches to PoU. If branch predictors are architecturally-visible, also
flushes branch predictors. Rt data: ignored"

soc_info.h
```diff
@@ -113,6 +113,8 @@ typedef struct {
 	uint32_t           rvbar_reg;    /* MMIO address of RVBARADDR0_L register */
 	const watchdog_info *watchdog;   /* Used for reset */
 	bool               sid_fix;      /* Use SID workaround (read via register) */
+	/* Use I$ workaround (disable I$ before first write to prevent stall thunk */
```
apritzel, line 116: I think it should end in "a stale thunk"

fel_lib.c
```diff
@@ -231,6 +232,33 @@ void aw_fel_execute(feldev_handle *dev, uint32_t offset)
 	aw_read_fel_status(dev);
 }
 
+static void aw_disable_icache(feldev_handle *dev)
+{
+	soc_info_t *soc_info = dev->soc_info;
+	uint32_t arm_code[] = {
```
apritzel, line 238: Can you please add a comment somewhere what this code does, exactly? For instance that you clear the SCTLR.I bit, then invalidate the I cache, and have the ISB to force the core to re-fetch instructions from there.

apritzel (commented):
Well, not too happy with that, in general, but if that fixes the problem for you, I guess we should take it, for now. Alternatively we could explore to invalidate the I$ after every code write. If some code snippet is put somewhere (towards the end of the thunk buffer?) early enough, we could branch to this every time before we want to execute new code.
One question remains, though: why does this code execute, but the other code you write doesn't? Is it because it's the first write to this location, so the corresponding I$ cache is still invalid?

Icenowy:
> Well, not too happy with that I-cache **hack**, in general, but if that fixes the problem for you, I guess we should take it, for now. Alternatively we could explore to invalidate the I$ after every code write. If some code snippet is put somewhere (towards the end of the thunk buffer?) early enough, we could branch to this every time before we want to execute new code. One question remains, though: why does this code execute, but the other code you write doesn't? Is it because it's the first write to this location, so the corresponding I$ cache is still invalid?

Yes, the offset inside 0x2000 of SRAM is the base of FES0 for BSP flashing, so it's promised to be able to be executed once.

Icenowy:
> > Well, not too happy with that I-cache **hack**, in general, but if that fixes the problem for you, I guess we should take it, for now. Alternatively we could explore to invalidate the aftereverycodewrite.Ifsomecodesnippetisputsomewhere(towardstheendofthethunkbuffer?)earlyenough,wecouldbranchtothiseverytimebeforewewanttoexecutenewcode.Onequestionremains,though:whydoesthiscodeexecute,buttheothercodeyouwritedoesn′t?Isitbecauseit′sthefirstwritetothislocation,sothecorrespondingI cache is still invalid?
> 
> Yes, the offset inside 0x2000 of SRAM is the base of FES0 for BSP flashing, so it's promised to be able to be executed once.

Oh sorry such new SoC use 0x8000 as FES offset, however as we need to know what SRAM is in use by BROM (to implement sunxi-fel spl), we should be still safe.

soc_info.c
```diff
@@ -392,6 +392,19 @@ soc_info_t soc_info_table[] = {
 		.sid_offset   = 0x200,
 		.rvbar_reg    = 0x08100040,
 		.watchdog     = &wd_h6_compat,
+	},{
+		.soc_id       = 0x1886, /* Allwinner V853 */
+		.name         = "V853",
+		/* Guessed values */
```
apritzel, line 398: I don't think this comment should be in. The values might not be optimal, I think the thunk and swap buffers should be pushed out, but this is not a big issue, and can be fixed later.

apritzel (approved):
Many thanks for the changes in the I$ code, that looks good now:
- I cannot verify that the I$ hack works, but I turned it around (setting SCTLR.I instead of clearing it), and I could see it going wrong, so I guess this is good now.
- The soc_info.c part looks alright, I will see if I can remove the (no longer needed) comment myself (did you allow maintainer edits?)
- The UART0 hello world GPIO changes look good. I will test on my T113 later today.
- I checked that the branch still works on a V5 (with support for that SoC added).

If hello world works on the T113, I will merge the PR.

soc_info.c
Icenowy: Yes I think I allowed maintainer edit.

BTW this comment is first here because when I write this code snippet, the user manual is still not available. So just removing it is okay now.

apritzel:
Indeed, on closer inspection the R528/T113 needs the same I cache hack as the V853. So I can confirm that this is required and works.
Playing around with the T113 also revealed that the watchdog moved, also requires a key in the upper half of the WDOG_MODE_REG. It seems the V853 is on the same page, so I fixed that up.
@Icenowy can you confirm that wdreset works for you on the V853 now? If yes, I will merge it.

Icenowy:
The wdreset function works flawlessly, thanks.

PR #179, opened 2022-07-29
Allwinner V5 and R528/T113 support

apritzel:
This adds support for the Allwinner V5, which is close to the H6, but with 32-bit cores.
Also basic support for the R528/T113, which is the same SoC as the RISC-V D1 (even using the same SoC ID), but with that RISC-V core disabled. Proper D1 support requires much more work to support non-ARM code uploads.
Also includes uart0 hello world support for both SoCs.

### DIFF
--- a/soc_info.c
+++ b/soc_info.c
@@ -422,6 +422,17 @@ soc_info_t soc_info_table[] = {
 		.icache_fix   = true,
 		.watchdog     = &wd_v853_compat,
 	},{
+		.soc_id       = 0x1721, /* Allwinner V5 */
+		.name         = "V5",
+		.spl_addr     = 0x20000,
+		.scratch_addr = 0x21000,
+		.thunk_addr   = 0x42200, .thunk_size = 0x200,
+		.swap_buffers = h6_sram_swap_buffers,
+		.sram_size    = 136 * 1024,
+		.sid_base     = 0x03006000,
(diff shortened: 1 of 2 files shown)


Icenowy:
Tested on Mango Pi MQ-R with T113.

Icenowy (commented):
BTW on D1s only 32K SRAM A1 is documented (maybe because it dropped DSP), but whether the DSP SRAM is available is up to testing.

Icenowy (approved):
V5 things are verified against the datasheet.

Icenowy:
> BTW on D1s only 32K SRAM A1 is documented (maybe because it dropped DSP), but whether the DSP SRAM is available is up to testing.

Someone says the DSP is still available, but just undocumented, so I think this should not be an issue now. (And I think D1 BROM assumes SRAM at a high position for its bss/stack)

apritzel:
Yes, I see the T113 BROM using the DSP SRAM, and can read and write to it as well. Since the D1(s) is explicitly not supported by this patch set anyway, we can address this later, if it's actually needed.

Icenowy:
Then I think I fully approve this PR.

Should I merge it?

apritzel:
> Should I merge it?

Yes, please! I guess there is little danger of regression here.

PR #180, opened 2022-07-29
Add A63 support

ChalesYu:
This adds support for the Allwinner A63, which is almost same as H6, but with different DRAM controller (maybe Designwire ?), and different interface.
The DRAM controller on A63 is closed to T7.
Also includes uart0 hello world support for this SoCs.

PR #184, opened 2022-11-27
uart0-helloworld-sdboot: update prebuilt blob

Icenowy:
The prebuilt blob of uart0-helloworld-sdboot isn't updated for a long time.

Update it with one built with 5cf618a21605 ("Merge pull request #180 from ChalesYu/a63-220729").

Signed-off-by: Icenowy Zheng <uwu@icenowy.me>


apritzel (approved):
The blob is identical to the one I just built from current HEAD. Tested on a T113s board.

PR #188, opened 2022-11-28
uart0-helloworld-sdboot: add support for F1C100s

apritzel:
The F1C100s series of SoCs has some subtle differences in its UART setup:
- The UART is at a previously unused address.
- The base clock is not APB2, but APB1.
- The input clock is not 24 MHz, but CPU clock / 2 / 2.
- The clock and reset gates are different bits at different addresses.

Add support for all those differences, tied to the F1C100s SoC ID.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,10 +171,10 @@ jtag-loop.elf: jtag-loop.c jtag-loop.lds
 	$(CROSS_CC) -g $(ARM_ELF_FLAGS) $< -nostdlib -o $@ -T jtag-loop.lds -Wl,-N
 
 fel-sdboot.elf: fel-sdboot.S fel-sdboot.lds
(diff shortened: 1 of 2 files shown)


Icenowy (approved):

Icenowy:
Code LGTM and tested.

However, should we add explicit -march=armv5te to the Makefile rules that generates these eGON files?

In addition, maybe the prebuilt blob can be updated in this PR too.

apritzel:
Yes, setting the target architecture explicitly is probably a good idea. If the Makefile ends up using Ubuntu's/Debian's arm-none-eabi-gcc compiler, we are compiling to armv4-te already anyways, so that should not create problems. I will test this later.
@Icenowy And yes, I just didn't merge your updated binary because I wanted to sneak in F1C100s support ;-)

PR #189, opened 2023-02-10
fel: Introduce 'sid-dump'

qianfan-Zhao:
A sample output from allwinner T3:

```console
$ ./sunxi-fel sid-dump
chipid          1300000c 02404700 79350400 5c761e0c
in              00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
ssk             00000000 00000000 00000000 00000000
thermal         08320834
ft_zone         00000000 00000000
tvout           010002b0 00f7029b 00f9029f 00fb02a5
rssk            00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
hdcp_hash       00000000 00000000 00000000 00000000
reserved        00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
                00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
                00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
                00000000 00000000 00000000 00000000
```

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -202,7 +202,7 @@ void *load_file(const char *name, size_t *size)
 		perror("Failed to open input file");
 		exit(1);
 	}
-	
+
 	while (true) {
 		size_t len = bufsize - offset;
 		size_t n = fread(buf+offset, 1, len, in);
@@ -216,7 +216,7 @@ void *load_file(const char *name, size_t *size)
 			exit(1);
 		}
 	}
-	if (size) 
+	if (size)
 		*size = offset;
 	if (in != stdin)
 		fclose(in);
@@ -438,6 +438,35 @@ void aw_fel_print_sid(feldev_handle *dev, bool force_workaround)
 	/* output SID in "xxxxxxxx:xxxxxxxx:xxxxxxxx:xxxxxxxx" format */
 	for (unsigned i = 0; i <= 3; i++)
 		printf("%08x%c", key[i], i < 3 ? ':' : '\n');
+}
+
+void aw_fel_dump_sid(feldev_handle *dev)
+{
+	uint32_t buffer[2048 / sizeof(uint32_t)]; /* total SID size is 2K */
+	soc_info_t *soc_info = dev->soc_info;
+
+	if (!soc_info->sid_base || !soc_info->sid_sections) {
+		printf("SID memory maps for your SoC (%s) are unknown.\n",
+			dev->soc_name);
+		return;
+	}
+
+	for (const sid_section *s = soc_info->sid_sections; s->name; s++) {
+		uint32_t count = s->size_bits / 32;
+
+		if (!fel_get_sid(dev, buffer, s->offset, count)) {
+			fprintf(stderr, "Read sid:%s failed\n", s->name);
+			return;
+		}
+
+		printf("%-15s", s->name);
+		for (uint32_t i = 0; i < count; i++) {
+			if (i > 0 && ((i % 8) == 0))
+				printf("\n%-15s", "");
+			printf(" %08x", buffer[i]);
+		}
+		putchar('\n');
+	}
 }
 
 void aw_enable_l2_cache(feldev_handle *dev, soc_info_t *soc_info)
@@ -1368,6 +1397,8 @@ int main(int argc, char **argv)
 			aw_fel_print_sid(handle, false);
 		} else if (strcmp(argv[1], "sid-registers") == 0) {
 			aw_fel_print_sid(handle, true); /* enforce register access */
+		} else if (strcmp(argv[1], "sid-dump") == 0) {
+			aw_fel_dump_sid(handle);
 		} else if (strcmp(argv[1], "write") == 0 && argc > 3) {
 			skip += 2 * file_upload(handle, 1, argc - 2, argv + 2,
 					pflag_active ? progress_bar : NULL);
--- a/fel_lib.c
+++ b/fel_lib.c
@@ -610,6 +610,15 @@ bool fel_get_sid_root_key(feldev_handle *dev, uint32_t *result,
 	return true;
 }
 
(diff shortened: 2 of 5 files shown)


fel.c
```diff
@@ -440,6 +440,43 @@ void aw_fel_print_sid(feldev_handle *dev, bool force_workaround)
...
+	if (!soc_info->sid_base || !soc_info->sid_sections) {
+		printf("SID memory maps for your SoC (%s) are unknown.\n",
+			dev->soc_name);
+		return;
+	}
+
+	for (const sid_section *s = soc_info->sid_sections; s->name; s++) {
+		uint32_t count = s->size_bits / 32;
+
+		if (count > sizeof(buffer) / sizeof(buffer[0])) {
```
apritzel, line 457: In common.h we have ARRAY_SIZE() for this. But in general this looks a bit odd to check this at all, since there is an upper limit (the 2KB SID storage size) and the input size comes from an internal struct, so we don't need to defend against user input.

fel_lib.c
```diff
@@ -610,6 +610,21 @@ bool fel_get_sid_root_key(feldev_handle *dev, uint32_t *result,
 	return true;
 }
 
+bool fel_get_sid(feldev_handle *dev, uint32_t *result, uint32_t offset,
+		 size_t count)
+{
+	if (!dev->soc_info->sid_base) {
```
apritzel, line 616: This check looks unnecessary, since we check this before calling this function.

fel_lib.c
```diff
@@ -610,6 +610,21 @@ bool fel_get_sid_root_key(feldev_handle *dev, uint32_t *result,
...
+		for (unsigned i = 0; i < count; i++)
+			result[i] = 0;
```
apritzel, line 618: First this would be a job for memset(). But why would we want to clear the result before we bail out? In fact I would expect a function to not touch a result buffer in case of an early error. Also the caller doesn't seem to need this, so remove that clearing part.

soc_info.c
```diff
@@ -507,3 +521,4 @@ void get_soc_name_from_id(soc_name_t buffer, uint32_t soc_id)
 	/* unknown SoC (or name string missing), use the hexadecimal ID */
 	snprintf(buffer, sizeof(soc_name_t) - 1, "0x%04X", soc_id);
 }
+
```
apritzel, line 524: stray extra line

apritzel (changes requested):
Hi Qianfan,
thanks for the patch, looks like a useful addition!
Can you please add some more text to the commit message, based on the text you used for the pull request?
And we would need the new command mentioned in the internal help text, ideally also in the manpage.
Some comments inside.

fel.c
qianfan-Zhao: done

fel_lib.c
qianfan-Zhao: OK, removed

fel_lib.c
qianfan-Zhao: Both the check for sid_base and the  clear function are removed now.

apritzel (approved):
Thanks for the quick changes, looks alright now. Tested on a Bananapi M2 Berry. I also tested with some A64 and H6 maps, will send them later.
I will let this sit for a few more days, in case people have comments.

fel_lib.c
```diff
@@ -610,6 +610,15 @@ bool fel_get_sid_root_key(feldev_handle *dev, uint32_t *result,
 	return true;
 }
 
+bool fel_get_sid(feldev_handle *dev, uint32_t *result, uint32_t offset,
+		 size_t count)
+{
+	fel_readl_n(dev, dev->soc_info->sid_base
+		    + dev->soc_info->sid_offset + offset, result, count);
```
smaeul, line 617: Should this use the register readout method on SoCs where that is required?

qianfan-Zhao: It's the better. I've considered it, but it's too hard to write assembly code.

apritzel: So it looks like `fel_get_sid_registers()` can be easily extended to read <n> registers from offset <x> instead of 4 registers from offset 0. Then we could just extend `fel_get_sid_root_key()` to allow parameters, and use that for both sid-dump and for `aw_fel_print_sid()`, avoiding a new function.
@qianfan-Zhao do you want to do this? If not, I can do it in a follow up patch.

qianfan-Zhao: @apritzel It is hard to me. Could you please submit some patchs after this pr merged?

apritzel: Ok, will do.

fel.c
```diff
@@ -440,6 +440,37 @@ void aw_fel_print_sid(feldev_handle *dev, bool force_workaround)
...
+		if (!fel_get_sid(dev, buffer, s->offset, count)) {
+			fprintf(stderr, "Read sid:%s failed\n", s->name);
+			return;
+		}
+
+		printf("%-16s", s->name);
+		for (uint32_t i = 0; i < count; i++) {
+			if (i > 0 && ((i % 8) == 0)) {
+				putchar('\n');
+				printf("%-16s", "");
```
apritzel, line 466: Just saw this on a final inspection: can you merge those two into one line:

    printf("\n%-16s", "");
Then you can get rid of the brackets as well.

fel.c
```diff
@@ -440,6 +440,37 @@ void aw_fel_print_sid(feldev_handle *dev, bool force_workaround)
...
+			}
+			printf("%08x ", buffer[i]);
```
apritzel, line 468: if you decrease the indent from 16 to 15 in the tow lines above, you can put the space here first, before the number, and avoid a trailing space on the last word.

apritzel (commented):
Just two minor things I noticed about the output formatting.

fel.c
qianfan-Zhao: done

fel.c
qianfan-Zhao: done

PR #190, opened 2023-02-23
Output a list of supported SoCs

apritzel:
The list of supported SoCs is ever growing, so it would be useful to check whether a particular build supports a certain SoC. Add an option to dump this list to the terminal.


fel.c
```diff
@@ -1286,6 +1287,7 @@ int main(int argc, char **argv)
 	bool uboot_autostart = false; /* flag for "uboot" command = U-Boot autostart */
 	bool pflag_active = false; /* -p switch, causing "write" to output progress */
 	bool device_list = false; /* -l switch, prints device list and exits */
+	bool list_socs = false; /* list all supported SoCs and exit */
```
paulkocialkowski, line 1290: You could call this `socs_list` to be consistent with the naming of the previous variable.

soc_info.c
```diff
@@ -509,6 +509,26 @@ soc_info_t *get_soc_info_from_version(struct aw_fel_version *buf)
...
+	if (prev == NULL)
+		return &soc_info_table[0];
+
+	for (soc = soc_info_table; soc->swap_buffers; soc++) {
+		if (soc == prev) {		/* is this the previous entry */
+			soc++;
+			if (soc->swap_buffers)	/* valid entry ? */
+				return soc;
+
+			return NULL;		/* we reached the end */
```
paulkocialkowski, line 525: I would invert both paths to make it clear that `!soc->swap_buffers` is a particular condition that needs to be handled but not the general case.

Also I think a general comment (with capital letter first and ending dot) would be a bit clearer to explain how this function works (which is nicely explained in the commit log) instead of broken-up single-line comments.

paulkocialkowski (changes requested):

fel.c
apritzel: That would be the other way around than the command line option, but fair enough, easy to change.

soc_info.c
apritzel: I am personally also a fan of sorting out the special cases first, but there are also merits to having positive if statements, to avoid negative logic.
But since you ask for it, I changed it, add added a comment, as suggested.

apritzel:
@paulkocialkowski, if you are fine with the changes and want to merge it, can you do a "Rebase and merge", to avoid the extra merge commit for just two patches? That makes a cleaner history, IMHO.

paulkocialkowski:
Thanks!

PR #191, opened 2023-02-23
Complete SID dump

apritzel:
Add some bits that were missing from the just merged sid-dump feature:
- support for the H3 register access work around
- maps for other SoCs
- documentation
@qianfan-Zhao @smaeul please have a look!

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -1273,6 +1273,9 @@ void usage(const char *cmd) {
 		"	echo-gauge \"some text\"		Update prompt/caption for gauge output\n"
 		"	ver[sion]			Show BROM version\n"
 		"	sid				Retrieve and output 128-bit SID key\n"
+		"	sid-registers			Retrieve and output 128-bit SID key,\n"
+		"					using the MMIO register read method\n"
+		"	sid-dump			Dump the content of all the SID eFuses\n"
 		"	clear address length		Clear memory\n"
 		"	fill address length value	Fill memory\n"
 		, cmd);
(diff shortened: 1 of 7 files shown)


Icenowy:
Looks good to me but have no time to test now.

fel.c
```diff
@@ -454,7 +454,7 @@ void aw_fel_dump_sid(feldev_handle *dev)
 	for (const sid_section *s = soc_info->sid_sections; s->name; s++) {
 		uint32_t count = s->size_bits / 32;
 
-		if (!fel_get_sid(dev, buffer, s->offset, count)) {
+		if (fel_read_sid(dev, buffer, s->offset, count * 4, false)) {
```
qianfan-Zhao, line 457: just a question: the `force_workaround` param was always false here, are we need to handle it as the same as 'sid' and 'sid-register'? such as add new command 'sid-register-dump' to make `force_workaround` as true?

apritzel: It's of course possible, but seems unnecessary to me: for every SoC other than the H3 the workaround is not necessary, and the memory provides the same content as the MMIO register view. The "sid-registers" command was just introduced to spot new SoCs with a similar problem. Adding yet another command might just confuse users.

soc_info.c
```diff
@@ -200,6 +200,66 @@ const sid_section r40_sid_maps[] = {
...
+	SID_SECTION("sck0_mask", 	0x1a0, 256),
+	SID_SECTION("sck1",	 	0x1c0, 256),
+	SID_SECTION("sck1_mask", 	0x1e0, 256),
+	SID_SECTION(NULL,	0,      0)
+};
+
+/* Placeholder for SoCs without a known SID map */
+const sid_section generic_2k_sid_maps[] = {
+	SID_SECTION("chipid",		0x00,  128),
+	SID_SECTION("unknonw",		0x10, 1920),
```
smaeul, line 260: typo: "unknown"

smaeul (approved):

PR #197, opened 2023-07-23
coverage dependency

PF4Public:
coverage should not start until testboards unpacked


apritzel (approved):
Indeed, calling `make coverage` without the `sunxi-boards/` directory populated fails, and this patch fixes it.

Issue #130, opened 2020-01-17
Does 'sunxi-fel spl' supports allwinner t3 soc?


qianfan-Zhao:
Next is the log, the soc is T3.

```
$ sudo sunxi-fel -l 
[sudo] password for x: 
USB device 003:005   Allwinner R40     12c00002:84404700:78868500:381f014f
$ sudo sunxi-fel -v spl boot0_sdcard_sun8iw11p1.bin
[sudo] password for x: 
Stack pointers: sp_irq=0x00002000, sp=0x00005E08
MMU is not enabled by BROM
usb_bulk_send() ERROR -7: Operation timed out
```

qianfan-Zhao:
I had tested the SRAM region, seems only 0xBC00 ~ 0xC000 can't write.

Test SRAM by using this scripts:
```
#!/bin/sh

CHECK_REGION_NUM=100
CHECK_REGION_ADDRESS_LIMIT=48128 # 0xBC00

# Write random number(size is 1K) to address($1) and then check if the FEL
# work fine.
check_region_1k_writeable () {
	local address=$(echo "obase=16;$1" | bc)
	local check_n=0

	printf "Check region(1K) 0x%04X " $1
	bin1k=$(mktemp)

	while [ ${check_n} -le ${CHECK_REGION_NUM} ] ; do
		local pass=0

		dd if=/dev/urandom of=${bin1k} bs=1K count=1 &> /dev/null
		if sunxi-fel write "0x${address}" ${bin1k} > /dev/null; then
			if sunxi-fel version > /dev/null; then
				let check_n++
				echo -n "."
				pass=1
			fi
		fi

		if [ ${pass} = 0 ] ; then
			echo "FAIL"
			return 1
		fi
	done

	echo "PASS"
}

# Check the SRAM region between 0x0000 and 0xC000 if it is writeabale.
address=0

while [ ${address} -le ${CHECK_REGION_ADDRESS_LIMIT} ] ; do
	# Skip the SRAM region in SP_IRQ(0x1c00~0x2000) and SP(0x5c00~0x7000)
	if [ ${address} -ge 7168 -a ${address} -lt 8192 ] || [ ${address} -ge 23552 -a ${address} -lt 28672 ] ; then
		let address+=1024
		continue
	fi

	check_region_1k_writeable ${address}
	if [ $? -ne 0 ] ; then exit 1; fi
	let address+=1024
done

```

Next is the result:

```
[...]

PR #204, opened 2024-01-02
fel: Add support for allwinner T7

qianfan-Zhao:
uart0-helloworld-sdcard can work with this patch:

```console
./sunxi-fel --verbose spl uart0-helloworld-sdboot.sunxi
Stack pointers: sp_irq=0x00022000, sp=0x00025E08
MMU is not enabled by BROM
=> Executing the SPL... done.
```

And next is the uart console's log:

```
Hello from Allwinner T7!
Returning back to FEL.
```

### DIFF
--- a/soc_info.c
+++ b/soc_info.c
@@ -285,6 +285,29 @@ static const sid_section h6_sid_maps[] = {
 	SID_SECTION(NULL,	0,      0)
 };
 
+static const sid_section t7_sid_maps[] = {
+	SID_SECTION("chipid",		 0x00, 128),
+	SID_SECTION("brom_config",	 0x10,  32),
+	SID_SECTION("thermal",		 0x14,  96),
+	SID_SECTION("tf_zone",		 0x20, 128),
+	SID_SECTION("oem",		 0x30, 128),
+	SID_SECTION("jtag-security",	 0x48,  32),
+	SID_SECTION("jtag-attr",	 0x4c,  32),
+	SID_SECTION("in",		 0x50, 192),
+	SID_SECTION("operator-id",	 0x68,  32),
+	SID_SECTION("id",		 0x6c,  32),
+	SID_SECTION("rotpk",		 0x70, 256),
+	SID_SECTION("ssk",		 0x90, 128),
+	SID_SECTION("rssk",		 0xa0, 256),
+	SID_SECTION("reserved",		 0xc0, 128),
+	SID_SECTION("ek_hash",	 	 0xd0, 128),
+	SID_SECTION("sn", 		 0xe0, 192),
+	SID_SECTION("nv1",	 	 0xf8,  32),
+	SID_SECTION("nv2",	 	 0xfc, 224),
+	SID_SECTION("reserved2", 	0x118, 320),
+	SID_SECTION(NULL,	0,      0)
+};
+
 /* Placeholder for SoCs without a known SID map */
 static const sid_section generic_2k_sid_maps[] = {
 	SID_SECTION("chipid",		0x00,  128),
@@ -435,7 +458,7 @@ soc_info_t soc_info_table[] = {
 		.sram_size    = 184 * 1024,
 		.sid_base     = 0x03006000,
 		.sid_offset   = 0x200,
-		.sid_sections = generic_2k_sid_maps,
+		.sid_sections = t7_sid_maps,
 		.watchdog     = &wd_h6_compat,
 	},{
 		.soc_id       = 0x1718, /* Allwinner H5 */
--- a/soc_info.c
+++ b/soc_info.c
@@ -113,6 +113,24 @@ sram_swap_buffers h6_sram_swap_buffers[] = {
 };
 
 /*
+ * T7 SRAM layout:
+ *
+ * SRAM A1: 0x0002_0000 - 0x0002_7fff,  32K contains stacks
+ * SRAM  C: 0x0002_8000 - 0x0004_ffff, 160K full access
+ * SRAM A2: 0x0010_0000 - 0x0010_3fff,  16K OpenRISC
+ *          0x0010_4000 - 0x0011_ffff, 112K full access
+ */
+sram_swap_buffers t7_sram_swap_buffers[] = {
+	/* 0x21C00-0x21FFF (IRQ stack) */
+	{ .buf1 = 0x21C00, .buf2 = 0x4e400, .size = 0x0400 },
+	/* 0x25C00-0x26FFF (Stack) */
+	{ .buf1 = 0x25C00, .buf2 = 0x4e800, .size = 0x1400 },
+	/* 0x27C00-0x27FFF (Something important) */
+	{ .buf1 = 0x27C00, .buf2 = 0x4fc00, .size = 0x0400 },
+	{ .size = 0 }  /* End of the table */
+};
+
+/*
  * V831 has 96KiB SRAM A1 at 0x20000 where the SPL has to be loaded to.
  * SRAM C is continuous with SRAM A1, and both SRAMs are tried to be used
  * by BROM. Memory space is allocated both from the start of SRAM A1 and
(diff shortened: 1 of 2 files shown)


soc_info.c
```diff
@@ -112,6 +112,23 @@ sram_swap_buffers h6_sram_swap_buffers[] = {
 	{ .size = 0 }  /* End of the table */
 };
 
+/*
+ * T7 SRAM layout:
+ *
+ * SRAM A1: 0x0002_0000 - 0x0002_7fff,  32K contains stacks
+ * SRAM  C: 0x0002_8000 - 0x0004_ffff, 160K full access
+ * SRAM A2: 0x0010_0000 - 0x0010_3fff,  16K unknown part, the data was changed
+ *                                          when read back
```
apritzel, line 121: Does the T7 have an OpenRISC management core? Then this will probably be its exception vector area, where only one word in every 256 byte region is writable, as seen in the other OpenRISC SoCs.

soc_info.c
```diff
@@ -112,6 +112,23 @@ sram_swap_buffers h6_sram_swap_buffers[] = {
...
+ *          0x0010_4000 - 0x0011_ffff, 112K full access
+ */
+sram_swap_buffers t7_sram_swap_buffers[] = {
+	/* 0x21C00-0x21FFF (IRQ stack) */
+	{ .buf1 = 0x21C00, .buf2 = 0x4d800, .size = 0x0400 },
```
apritzel, line 126: Do we need such a large stack? Do you know the stack pointer value?

soc_info.c
```diff
@@ -407,6 +424,18 @@ soc_info_t soc_info_table[] = {
 		.sid_base     = 0x01C23800,
 		.sid_sections = generic_2k_sid_maps,
 		.watchdog     = &wd_h3_compat,
+	},{
+		.soc_id       = 0x1708, /* Allwinner T7 */
+		.name         = "T7",
+		.spl_addr     = 0x20000,
+		.scratch_addr = 0x21000,
+		.thunk_addr   = 0x4d200, .thunk_size = 0x200,
```
apritzel, line 432: There seems to be a gap between this thunk area and the swap buffers. Is that intentional? Can we move this area here up to end at the swap buffer's start?

uart0-helloworld-sdboot.c
```diff
@@ -513,6 +515,10 @@ void gpio_init(void)
 		sunxi_gpio_set_cfgpin(SUNXI_GPE(2), SUN8I_R528_GPE_UART0);
 		sunxi_gpio_set_cfgpin(SUNXI_GPE(3), SUN8I_R528_GPE_UART0);
 		sunxi_gpio_set_pull(SUNXI_GPE(3), SUNXI_GPIO_PULL_UP);
+	} else if (soc_is_t7()) {
```
apritzel, line 518: That looks to be the same as the A64. Can you please double check and then just add soc_is_t7() to the existing A64 check?

apritzel (changes requested):
That looks mostly good, thanks for sending this! Some comments inside.

soc_info.c
qianfan-Zhao: I had not find the OpenRISC core in the datasheet.  I had write 16KB zero data to this section and then readback, next is the new data:

```
$ hexdump -C 2.bin
00000000  00 00 00 00 00 00 00 15  00 00 00 00 00 00 00 00  |................|
00000010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000100  00 00 00 00 00 00 00 15  00 00 00 00 00 00 00 00  |................|
00000110  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000200  00 00 00 00 00 00 00 15  00 00 00 00 00 00 00 00  |................|
00000210  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
...
```

Maybe there only one word in every 256 byte region is not writable. And I don't known what the means of 0x15.

soc_info.c
qianfan-Zhao: I had write a tool like #130, write 1K random data everytime and then readback, when I write data to 0x21c00 the bootROM will hang. So I think this section size is 1K, maybe reducing the amount of data written each time will get a  more accurate results, but that's not important.

when startting u-boot-spl in verbose mode, it will print the stack pointer value:

```
./sunxi-fel --verbose spl uart0-helloworld-sdboot.sunxi
Stack pointers: sp_irq=0x00022000, sp=0x00025E08
MMU is not enabled by BROM
=> Executing the SPL... done.
```

the *sp_irq* value can match it.

soc_info.c
qianfan-Zhao: No enough reason. The swap buffer is started at 0x4d800 and I make the thunk address before swap buffer. And reserve some space if we need extended the thunk_size in the future.

uart0-helloworld-sdboot.c
qianfan-Zhao: Sure.

soc_info.c
apritzel: Yes,that's exactly the OpenRISC exception vector pattern we see in the other SoCs, compare here: https://linux-sunxi.org/A64/Memory_map

soc_info.c
apritzel: Thanks for the stack pointers! Yes, the IRQ stack looks correct, both the size and location. I was wondering about the normal stack, though. Can you dump the BootROM somewhere? The initial FEL stack pointers are pretty easy to find in the disassembly, then we could reserve 5K down from there, to be on the safe side, and make sure the 0x25e08 is in that region.

soc_info.c
apritzel: All other SoCs use 0x200, and the current code uses much less than that, If we would need more, we would need to change the value for all other SoCs, too. So there is no reason to go beyond 512 bytes, or to reserve space already.

soc_info.c
apritzel: To elaborate: the first word is writable: that receives the branch instruction containing the target address of the exception handler. The 0x15000000 is the hardcoded NOP needed for the delay slot. The rest of each 256 byte region is not implemented and hardwired to 0.

qianfan-Zhao:
Hi @apritzel , the attachment is the N-BROM dump with https://github.com/DrSchottky/Allwinner-BROM-dumper:

```console
$ ./dump 0 0xc000 T7-N-BRROM.bin 
BROM magic (eGON.BRM) found!
BROM header starts at 0x3000
BROM header size is 36
BROM boot version is 1100
BROM eGON is 1100
BROM platform is 1633
BROM dumped successfully!
```

The BROM set SP to 0x27ffc in the address 0x00003294 and I am not a good hacker for arm asm, could you please check it again?

[T7-N-BROM.bin.gz](https://github.com/linux-sunxi/sunxi-tools/files/13856991/T7-N-BROM.bin.gz)

apritzel:
Many thanks for the NBROM dump, that's very helpful!
So the FEL entry point is at 0x20, from there it branches to 0x70, and sets up the two stack pointers very early: the IRQ SP to 0x22000 (stored at 0x2ec), and the SVC SP to 0x27000 (stored at 0x2f0). Since the BROM also contains code for loading from MMC or NAND or SPI flash, there are more routines setting stack pointers, so I think the value you found belongs to other routines?
So with that it looks like just 5K would be enough. Can you check whether the area between 0x27000 and 0x28000 contains any data or is written to by the BROM during FEL operation? Then I guess we can reduce the stack backup buffer to 5K.

qianfan-Zhao:
@apritzel you are right that the SVC sp is started at 0x27000. We can touch the data between 0x27000 and 0x27c00.

```console
$ ./check.sh 0x27000 0x27fff  
Check region(1K) 0x27000 ...........
Check region(1K) 0x27400 ...........
Check region(1K) 0x27800 ...........
Check region(1K) 0x27C00 usb_bulk_send() ERROR -7: Operation timed out
usb_bulk_send() ERROR -7: Operation timed out
sha1sum doesn't match
```

The swap buffer of T7 is much more like H6, 0x27c00 is someting important and we also can not touch it.

apritzel (approved):
Awesome, many thanks for checking all the regions, and for the changes! Looks all good now, I also checked the SID maps for overlaps.

PR #206, opened 2024-01-28
fel: sid: add sid-writel command

gamelaster:
This pull request introduces the capability to write into SID eFuses. However, it is worth noting that this is a risky operation that could potentially brick the SoC. As a precaution, this code will be disabled by default, and packagers should not ship binaries with this functionality enabled.

Before merging, those stuff needs to be solved:

- What will be the best way of enabling this feature? Makefile argument? Or uncommenting some define?
- Currently, only a 32-bit value write command is implemented. Although, for purposes such as writing root private key, writing a file command would be handy, although, this also can easily lead to bricked SoC in case the user uses it wrongly. The idea I had on sunxi-linux IRC was, that every write of file into SID, will print where and how many bytes will be written to SID, and the user will need to confirm to proceed.
Example: `14 bytes will be written to SID, offset 0x18, thus this range of SID will be burned <0x18, 0x26>. Do you want to continue?`

Looking forward for suggestions!

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -469,6 +469,19 @@ void aw_fel_dump_sid(feldev_handle *dev)
 	}
 }
 
+void aw_fel_writel_sid(feldev_handle *dev, uint32_t offset, uint32_t value)
+{
+	soc_info_t *soc_info = dev->soc_info;
+
+	if (!soc_info->sid_base || !soc_info->sid_sections) {
+		printf("SID memory maps for your SoC (%s) are unknown.\n",
+			dev->soc_name);
+		return;
+	}
+
+	fel_write_sid(dev, &value, offset, 4);
+}
+
 void aw_enable_l2_cache(feldev_handle *dev, soc_info_t *soc_info)
 {
 	uint32_t arm_code[] = {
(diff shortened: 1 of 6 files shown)


qianfan-Zhao:
Hi:

I had a idea about this but have not complete it. 
We had introduced a subcommand in #189, which can print the sid registers for better human reading. Next is the example:

```
$ ./sunxi-fel sid-dump
chipid          1300000c 02404700 79350400 5c761e0c
in              00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
ssk             00000000 00000000 00000000 00000000
thermal         08320834
ft_zone         00000000 00000000
tvout           010002b0 00f7029b 00f9029f 00fb02a5
rssk            00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
hdcp_hash       00000000 00000000 00000000 00000000
reserved        00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
                00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
                00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
                00000000 00000000 00000000 00000000
```

We defined the sid maps layout in the code: https://github.com/linux-sunxi/sunxi-tools/blob/master/soc_info.c#L221, we known the section's size and also can define writeable properity in the this section.

So it's better for this command line:

```
$ sunxi-fel sid-write section-name 'aabbcc'
```

What's yours idea?

PR #210, opened 2024-08-27
A523 FEL fixes

apritzel:
While playing with a proper A523 (T527, really) devboard, I noticed some issues with the FEL support: We need to disable the I-cache, to allow uploading *and* executing code via FEL.
Also the load and buffer addresses are not ideal: the BROM actually loads any MMC payload somewhere else, so we need to follow suit here to keep a single SPL image, for loading via both SD card/eMMC/SPI and via FEL.


jernejsk (approved):
Confirmed as needed, LGTM.

PR #211, opened 2024-09-23
Rework uart0-helloworld-sdboot to use SoC table

apritzel:
Over time the seemingly simple uart0-helloworld-sdboot program became a bit of a mess: new SoCs tend to mix and match various properties, and people adding support for them needed to find the right group to put them into. This lead to long `if (soc_is_xxx() || soc_is_yyy() ...)` chains, which are not only hardly readable, but made any changes quite error-prone.

This series reworks the whole design, by adding a SoC table (inspired by soc_info.c), where we hold various values for each SoC: among the SoC ID and its name things like the UART base address, the clock base, and the UART0 pins.
This not only simplifies the whole code, but makes adding support for new SoCs a breeze: as a bonus the last three patches add support for the A83T, the A80 and the A523. And even though this table takes about 1KB of RAM (for all the 23 supported SoCs), the whole binary got actually much smaller:
```
   text    data     bss     dec     hex filename
   4207       0      24    4231    1087 uart0-helloworld-sdboot.elf (before)
   2441       0      20    2461     99d uart0-helloworld-sdboot.elf (after)
```
Please review, I split the patches up to help with that process.


jernejsk (approved):

paulkocialkowski:
Great work, I've tried this on a few platforms (including A80) and it works nicely.

I've also been using this to rework the `jtag-loop` implementation to work on all SoCs (it essentially remaps the sdcard pins to jtag), which I can submit for review once it's cleaned-up and doesn't duplicate the code. So I might end-up moving this code to a common helper file.

PR #212, opened 2024-10-12
Fix fel-remotefunc-compiler execution on newer Ruby versions

iscle:
It was using a deprecated (now deleted) function to check file existence.

### DIFF
--- a/fel-remotefunc-compiler.rb
+++ b/fel-remotefunc-compiler.rb
@@ -48,7 +48,7 @@ def tool_exists(tool_name)
 end
 
 def parse_stack_usage(filename)
-  return unless File.exists?(filename)
+  return unless File.exist?(filename)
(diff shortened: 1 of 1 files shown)


paulkocialkowski (commented):
Does the new call also work with older Ruby versions? If not it would be good to be able to support both cases.

iscle:
> Does the new call also work with older Ruby versions? 

Yes! The "exist" method has always existed in Ruby, so there shouldn't be any issues :)

Here's the link to the documentation of the oldest available Ruby version, with the method already present: https://ruby-doc.org/core-1.8.6/File.html#method-c-exist-3F

apritzel:
Yes, can confirm that it works here, and `.exist` is there from very early on, apparently. `.exists` on the other hand was already marked obsolete in that 20 year old v1.8 documentation.

PR #213, opened 2024-12-13
Add B288 support

Soupborsh:
Notably this makes `sunxi-fel spl` , `sunxi-fel uboot` and  `sunxi-fel wdreset` to work, tested on PB618

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,7 +186,7 @@ phoenix_info: phoenix_info.c
 	$(SUNXI_BOOT_IMAGE) $< $@
 
 ARM_ELF_FLAGS = -Os -marm -fpic -Wall
-ARM_ELF_FLAGS += -fno-common -fno-builtin -ffreestanding -nostdinc -fno-strict-aliasing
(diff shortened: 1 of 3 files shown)


apritzel:
Please squash this into one patch, and add some commit message, briefly explaining what this SoC is about and how you came to the values.
Also this looks like a 1:1 copy of the H3 data, and without further confirmation or a user manual I am a bit doubtful this is really correct (for instance the SRAM size).
So can you please say what you have tested and how?
Does the SID dump work? Do you really need the .sid_fix?
Some parts can be deduced from the BootROM, can you dump this and upload somewhere?

    $ sunxi-fel read 0xffff0000 65536 brom_b288.bin

Also https://github.com/smaeul/sunxi-blobs/ contains a collection of dumped BootROMs, so you could post your version as a PR there.

ValdikSS:
> Some parts can be deduced from the BootROM, can you dump this and upload somewhere?

https://github.com/smaeul/sunxi-blobs/pull/6

PR #216, opened 2025-02-09
fel-lib: Fix parsing scratchpad field

jernejsk:
Field has 32 bits, not 16, as it is currently handled. This fixes scratchpad address print when using "sunxi-fel ver" command on A523. Before it printed "0x1500" and afterwards "0x61500", which makes much more sense and it was verified by BROM disassembly.


apritzel (approved):
Looks good to me, it's a uint32_t anyway, and indeed there are more than 16 significant bits on the A523.

PR #218, opened 2025-03-17
T113S SPI support

dron0gus:
This adds SPI support for R528/T113-S3.
T113 has different SPI, GPIO and CCM (CCU - clock control unit) base addresses. Also it has different step between gpio banks, so now `gpio_base()` returns base for given `port_num`.
Tested on LCPI-PC-F133-T113-D1S-V1.3 board with Windbond W25Q128JVSIQ SPI flash soldered.
```
$ sudo ./sunxi-fel spiflash-info
Manufacturer: Winbond (EFh), model: 40h, size: 16777216 bytes.
```
Write and read are also tested with some random file and md5sum check of initial and readback files.

### DIFF
--- a/fel-spiflash.c
+++ b/fel-spiflash.c
@@ -154,6 +154,8 @@ static uint32_t spi_base(feldev_handle *dev)
 	case 0x1728: /* H6 */
 	case 0x1823: /* H616 */
 		return 0x05010000;
+	case 0x1859: /* D1/D1s/R528/T113-S3 */
+		return 0x04025000;
 	default:
 		return 0x01C68000;
 	}
@@ -207,6 +209,7 @@ static bool soc_is_h6_style(feldev_handle *dev)
 	case 0x1817: /* V831 */
 	case 0x1728: /* H6 */
 	case 0x1823: /* H616 */
+	case 0x1859: /* D1/D1s/R528/T113-S3 */
 		return true;
 	default:
 		return false;
--- a/fel-spiflash.c
+++ b/fel-spiflash.c
@@ -90,6 +90,7 @@ void fel_writel(feldev_handle *dev, uint32_t addr, uint32_t val);
 #define SUNIV_GPC_SPI0              (2)
 #define SUNXI_GPC_SPI0              (3)
 #define SUN50I_GPC_SPI0             (4)
+#define SUN8I_GPC_SPI0				(2)
 
 #define SUN4I_CTL_ENABLE            (1 << 0)
 #define SUN4I_CTL_MASTER            (1 << 1)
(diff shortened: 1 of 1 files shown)


fel-spiflash.c
```diff
@@ -174,11 +176,11 @@ static uint32_t ccm_base(feldev_handle *dev)
 static void gpio_set_cfgpin(feldev_handle *dev, int port_num, int pin_num,
 			    int val)
 {
-	uint32_t port_base = gpio_base(dev) + port_num * 0x24;
+	uint32_t port_base = gpio_base(dev, port_num);
 	uint32_t cfg_reg   = port_base + 4 * (pin_num / 8);
 	uint32_t pin_idx   = pin_num % 8;
 	uint32_t x = readl(cfg_reg);
-	x &= ~(0x7 << (pin_idx * 4));
+	x &= ~(0xf << (pin_idx * 4));
```
paulkocialkowski, line 186: You could mention in the commit log that the field now uses 4 bits. It was discussed in #208
Also you might want to include changes from this PR as it seems stalled currently.

fel-spiflash.c
```diff
@@ -156,6 +156,18 @@ static uint32_t spi_base(feldev_handle *dev)
 	}
 }
 
+static uint32_t ccm_base(feldev_handle *dev)
+{
+	soc_info_t *soc_info = dev->soc_info;
+	switch (soc_info->soc_id) {
+	case 0x1859: /* D1/D1s/R528/T113-S3 */
+		// CCU
```
paulkocialkowski, line 169: This comment could be removed, also it's not the same comment style as the rest of the file (`/* */`).

fel-spiflash.c
```diff
@@ -90,6 +90,7 @@ void fel_writel(feldev_handle *dev, uint32_t addr, uint32_t val);
 #define SUNIV_GPC_SPI0              (2)
 #define SUNXI_GPC_SPI0              (3)
 #define SUN50I_GPC_SPI0             (4)
+#define SUN8I_GPC_SPI0				(2)
```
paulkocialkowski, line 93: Unfortunately the original author of the file aligned with spaces, but please stick to it.

paulkocialkowski (changes requested):
Thanks for your work! See some specific comments inline.
You might want to include the work from #208 since you're touching the same areas.

Also please use a "fel-spiflash: " prefix for the commit messages (which makes it easier to see which part of sunxi-tools each commits affects) and generally use an action-oriented wording such as "Add support for T113-S3 SPI" rather than a descriptive title.

Other than the requested changes it all looks good, nice work!

PR #220, opened 2025-03-19
Allwinner A133 sunxi-fel and UART support

apritzel:
This adds support for the A133 SoC, with the A100 and the R818 sharing the same die, and being apparently fully compatible. Nothing special here, the SoC is quite similar to the already supported H6, so not many changes this time.


loki666:
tested on a MagicX Mini Zero28 sporting a A133P.
works fine

paulkocialkowski (approved):

PR #221, opened 2025-04-16
Move uart0-helloworld-sdboot common code to bare-metal and add JTAG loop support

paulkocialkowski:
This makes it easier to develop bare-metal programs and adds support for muxing JTAG on the SD card pins for most SoCs.


apritzel (approved):

apritzel (changes requested):
sorry, was just looking at the first patch, didn't mean to approve the whole series just yet ...

bare-metal.c
apritzel, line 1: Looks like a legit copy of the original file, stripped down to the common functions, and slight adjustments like removing the const modifiers. Also doesn't change the code size.
So all good.

bare-metal.h
```diff
@@ -186,15 +186,27 @@ enum sunxi_gpio_number {
...
-	char	soc_name[10];
-	u32	pio_base;
-	u32	ccu_base;
-	u32	sram_a1_base;
-	u32	uart0_base;
-	u16	uart0_tx_pin;
-	u8	uart0_pinmux;
+	u16	id;
+	char	name[10];
 	u8	flags;
```
apritzel, line 192: Unfortunately this breaks the neatly aligned struct, making it 36 bytes per SoC instead of 32 bytes due to padding. So lets light a candle for that, but then move on, and future additions would have done this anyway.
Grows that binary by 68 bytes, but I guess we can live with that.

bare-metal.c
```diff
@@ -104,6 +114,8 @@ static const struct soc_info soc_table[] = {
 		.ccu	= { AW_CCM_BASE },
 		.sram	= { SRAM_A1_ADDR_0 },
 		.uart0	= { SUNXI_UART0_BASE, SUNXI_GPB(22), MUX_2 },
+		.jtag	= { SUNXI_GPF(5), SUNXI_GPF(3), SUNXI_GPF(1),
```
apritzel: So I appreciate the idea of documenting the oddness of just the A20 using a slightly different JTAG pin assignment, but it looks to be completely irrelevant for the jtag tool, as it just applies the same mux to all pins listed. This change increases the binary by quite a sizeable chunk (364 bytes or 15%), and for something that uart0-helloworld-sdboot doesn't even use.
The pins are always the same (as are the UART pins on PortF, so no coincidence), so we don't need to store them at all, and can just hardcode them. We do this already for the UART-on-PortF pins. This would just leave the pinmux to store, which still enlarges the binary, but not by that much.

jtag-loop.c
apritzel, line 1: That looks like a very nice change, not only re-using the existing code, but also actually enabling it for all SoCs instead of just the handful that it was covering before. Like it.

bare-metal.c
```diff
@@ -609,6 +609,24 @@ void uart0_puts(const char *s)
 	}
 }
 
+void uart0_puthex(unsigned long hex)
```
apritzel: Looks alright, I had a similar version in my own baremetal toolbox. I found it actually very handy to have the possibility to add an EOL after the number, so maybe add a ``bool add_eol`` here? That allows for this pattern:

    uart0_puts("SCTLR: 0x");
    uart0_puthex(reg, EOL);

What do you think? What were your use cases?

apritzel (changes requested):
So I think the changes look alright, and indeed allow for easier additions, particularly for other bare metal applications. I would just ask for the JTAG pin numbers to be removed, see the comment on the respective commit for more details.

bare-metal.h
paulkocialkowski: Yep. Is there particular interest about the alignment?

bare-metal.c
paulkocialkowski: Yes I understand the argument that it's a bit redundant and could work out with just the mux number and pins hardcoded. However I don't really see why it's a problem to increase the binary size. Is there a limit to what fel load can handle?

jtag-loop.c
paulkocialkowski: And thanks for reviewing :)

bare-metal.c
paulkocialkowski: Yes my use case was just printing some registers for debug, which is indeed a lot more readable with a newline after the value. I can add an argument for it.

bare-metal.c
ssvb: > However I don't really see why it's a problem to increase the binary size. Is there a limit to what fel load can handle?

I would worry more about the source code LoC count and overall complexity. This uart0-helloworld-sdboot code was intended as a simple and easy to comprehend example in its entirety, so that people could learn from it and understand what's the minimum amount of required non-overengineered code to get the device up and running. Rather than some sort of an opaque complicated magic framework hidden behind an easy to use API (which might be nice for other purpuses, but has poor educational value here). For comparison, the u-boot SPL is insanely overengineered and difficult to comprehend by the beginners.

paulkocialkowski: I understand the initial intent but there is definitely a need for something that works on all supported platforms without duplication and the JTAG use-case (which is no longer a demo but something actually practically useful) shows that pretty well. I think we're still on the good side of "overengineered" :)

bare-metal.c
paulkocialkowski: Mhh I just realized I actually ditched the entry for A133 when doing this...

Makefile
```diff
@@ -168,22 +168,22 @@ ARM_ELF_FLAGS += -mno-thumb-interwork -fno-stack-protector -fno-toplevel-reorder
 ARM_ELF_FLAGS += -Wstrict-prototypes -Wno-format-nonliteral -Wno-format-security
 
 jtag-loop.elf: jtag-loop.c jtag-loop.lds
-	$(CROSS_CC) -g $(ARM_ELF_FLAGS) $< -nostdlib -o $@ -T jtag-loop.lds -Wl,-N
+	$(CROSS_CC) -march=armv5te -g $(ARM_ELF_FLAGS) $< -nostdlib -o $@ -T $(lastword $^) -Wl,-N
```
apritzel: While not wrong, the -march change is a bit surprising here. Can you move that to the but-last patch, that touches that line anyway?

bare-metal.c
apritzel: @ssvb: I agree with @paulkocialkowski here: it's only simple until it isn't anymore. And while uart0-helloworld might have been an example with some educational value a decade ago, when people talked about the "sunxi chip" and meant the A10, the addition of more SoCs (25 now!) spoiled this already years ago, and it actually became less readable and I'd argue almost unsuitable for learners. And enabling jtag-loop on modern SoCs is a worthwhile reason to refactor this into some kind of library.
There is always the old code in the git history if people are really looking for the simplest possible code - or we create a separate repo for that.

bare-metal.h
apritzel: Not really, just to stupid desire to have things everything neat and tidy. I just want to keep an eye on the binary size, but this amount of grow is certainly acceptable.

bare-metal.c
```diff
@@ -53,64 +53,232 @@
...
+		.flags	= FLAG_VAR1 | FLAG_UART_ON_PORTF,
+		.pio	= { SUNXI_PIO_BASE },
+		.ccu	= { AW_CCM_BASE },
+		.sram	= { SRAM_A1_ADDR_0 },
+		.uart0	= { SUNXI_UART0_BASE, SUNXI_GPB(19), MUX_2 },
+	},
+	{
+		.id	= 0x1633,
+		.name	= "A31/A31s",
+		.flags	= FLAG_VAR1,
```
apritzel: This seems wrong, there is only one SoC with that SoC ID, so need for a variant. The original struct didn't have this.

apritzel (approved):
I fixed the two small things I mentioned, the code looks good now to me.

PR #223, opened 2025-08-14
Add FEL support for Allwinner SoC A733

dlan17:
the PR tries to add FEL support for A733, tested on Radxa Cubie A7A board
it's capable of booting uboot spl image..

still need some help to check if the commit message is accurate..

the spl_address is coded to 0x52000, as in my case I've tried, the <0x51000 isn't working..

### DIFF
--- a/soc_info.c
+++ b/soc_info.c
@@ -189,6 +189,12 @@ sram_swap_buffers a523_sram_swap_buffers[] = {
 	{ .buf1 = 0x45000, .buf2 = 0x40200, .size = 0x0400 },
 	{ .size = 0 }  /* End of the table */
 };
+
+sram_swap_buffers a733_sram_swap_buffers[] = {
+	{ .buf1 = 0x47000, .buf2 = 0x72400, .size = 0x1C00 },
+	{ .size = 0 }  /* End of the table */
+};
+
 /*
  * Some SoCs put both stacks, BSS and data segments at the end of a comparably
  * large SRAM, so we don't need to move anything around.
@@ -642,6 +648,20 @@ soc_info_t soc_info_table[] = {
 		.rvbar_reg    = 0x08100040,
 		.needs_smc_workaround_if_zero_word_at_addr = 0x100004,
 		.watchdog     = &wd_h6_compat,
+	},{
+		.soc_id       = 0x1903, /* Allwinner A733 */
+		.name         = "A733",
+		.spl_addr     = 0x47000,
+		.scratch_addr = 0x71000,
+		.thunk_addr   = 0x72000, .thunk_size = 0x200,
+		.swap_buffers = a733_sram_swap_buffers,
+		.sram_size    = 180 * 1024,
+		.sid_base     = 0x03006000,
+		.sid_offset   = 0x200,
+		.sid_sections = generic_2k_sid_maps,
+		.rvbar_reg    = 0x08001004,
+		.icache_fix   = true,
+		.watchdog     = &wd_a523_compat,
 	},{
 		.swap_buffers = NULL /* End of the table */
 	}
--- a/uart0-helloworld-sdboot.c
+++ b/uart0-helloworld-sdboot.c
@@ -80,6 +80,7 @@ typedef unsigned char u8;
 #define R329_UART0_BASE		0x02500000
 #define R329_PIO_BASE		0x02000400
 #define R329_CCM_BASE		0x02001000
+#define A733_CCM_BASE		0x02002000
 
 #define V853_PIO_BASE		0x02000000
 
@@ -175,8 +176,11 @@ enum sunxi_gpio_number {
 #define FLAG_NEW_CLOCK		BIT(3)
 #define FLAG_UART_ON_APB1	BIT(4)
 #define FLAG_A80_CLOCK		BIT(5)
+#define FLAG_A733_GPIO		BIT(6)
+#define FLAG_A733_CLOCK		BIT(7)
 
 #define FLAG_NCAT2		FLAG_NEW_GPIO | FLAG_NEW_CLOCK
+#define FLAG_NCAT3		FLAG_A733_GPIO | FLAG_A733_CLOCK
 
 static const struct soc_info {
 	u16	soc_id;
@@ -239,6 +243,8 @@ static const struct soc_info {
 		R329_UART0_BASE, SUNXI_GPH(9), MUX_5, FLAG_NCAT2 },
 	{ 0x1890, "A523", V853_PIO_BASE, R329_CCM_BASE, SRAM_A1_ADDR_20000,
 		R329_UART0_BASE, SUNXI_GPB(9), MUX_2, FLAG_NCAT2 },
+	{ 0x1903, "A733", V853_PIO_BASE, A733_CCM_BASE, SRAM_A1_ADDR_20000,
+		R329_UART0_BASE, SUNXI_GPB(9), MUX_2, FLAG_NCAT3 },
 };
 
 #define ARRAY_SIZE(x) (sizeof(x) / sizeof(x[0]))
@@ -259,6 +265,7 @@ static const struct soc_info *find_soc_info(int soc_id, int variant)
 }
 
 static u32 pio_base;
+static u32 pio_bank_a_offset;
 static u32 pio_bank_size, pio_dat_off, pio_pull_off;
 
 static int sunxi_gpio_set_cfgpin(u32 pin, u32 val)
@@ -411,6 +418,9 @@ static void clock_init_uart(const struct soc_info *soc)
 	if (soc->flags & FLAG_NEW_CLOCK) {
 		set_wbit(soc->ccu_base + 0x90c,
 			 0x10001 << (CONFIG_CONS_INDEX - 1));
+	} else if (soc->flags & FLAG_A733_CLOCK) {
+		set_wbit(soc->ccu_base + 0xe00 + (CONFIG_CONS_INDEX - 1) * 4,
+			 0x10001);
 	} else {
 		int bit = 16 + CONFIG_CONS_INDEX - 1;
 		int gate_ofs = 0x06c;
@@ -439,19 +449,25 @@ static void clock_init_uart(const struct soc_info *soc)
 
 static void gpio_init(const struct soc_info *soc)
 {
-	pio_base = soc->pio_base;
-
 	if (soc->flags & FLAG_NEW_GPIO) {
 		/* GPIO V2 */
 		pio_bank_size = 0x30;
 		pio_dat_off = 0x10;
 		pio_pull_off = 0x24;
+	} else if (soc->flags & FLAG_A733_GPIO) {
+		/* Bank A Offset = 0x80 */
+		pio_bank_a_offset = 0x80;
+		pio_bank_size = 0x80;
+		pio_dat_off = 0x10;
+		pio_pull_off = 0x30;
 	} else {
 		/* GPIO V1 */
 		pio_bank_size = 0x24;
 		pio_dat_off = 0x10;
 		pio_pull_off = 0x1c;
 	}
+
+	pio_base = soc->pio_base + pio_bank_a_offset;
 
 	if (soc->flags & FLAG_UART_ON_PORTF) {
 		/* Disable normal UART0 pins to avoid conflict */


Icenowy:
Is the WIP U-Boot port available anywhere?

Icenowy:
And is there any BROM dump of this SoC?

dlan17:
> Is the WIP U-Boot port available anywhere?

have a local repo, but not sorted out yet,
the main challenge is to bring up DDR..

dlan17:
> And is there any BROM dump of this SoC?

no idea of what's the aim to do the dump, not really needed of this from my side..

soc_info.c
```diff
@@ -189,6 +189,12 @@ sram_swap_buffers a523_sram_swap_buffers[] = {
 	{ .buf1 = 0x45000, .buf2 = 0x40200, .size = 0x0400 },
 	{ .size = 0 }  /* End of the table */
 };
+
+sram_swap_buffers a733_sram_swap_buffers[] = {
+	{ .buf1 = 0x48800, .buf2 = 0x72000, .size = 0x0400 },
```
apritzel: Can you double check that the area at 0x72000 is not used by the BROM?
Try to write something to that area, using "sunxi-fel write", then try to read it back (using the "read") command. This should work without using the swap buffers, and this way you can check if the BROM uses that area. Feel free to poke around nearby, it would be good if you could actually find something that the BROM changes, and the confirm that it's NOT between 0x72000 and 0x72400.
Another way would be to dump the BootROM, then check the disassembly for addresses used for buffers and stack.

soc_info.c
```diff
@@ -642,6 +648,20 @@ soc_info_t soc_info_table[] = {
 		.rvbar_reg    = 0x08100040,
 		.needs_smc_workaround_if_zero_word_at_addr = 0x100004,
 		.watchdog     = &wd_h6_compat,
+	},{
+		.soc_id       = 0x1903, /* Allwinner A733 */
+		.name         = "A733",
+		.spl_addr     = 0x52000,
```
apritzel: So this should be 0x47000, to match the load address used by the BROM when loading from SD cards (and other boot media). Making this the same is crucial to be able to use the same SPL binary for FEL and boot media, since we need a fixed link address for the SPL (since we cannot use PIE).
And we should double check that it's really 0x47000, and not 0x44000. Again the BROM disassembly would help.

soc_info.c
```diff
@@ -642,6 +648,20 @@ soc_info_t soc_info_table[] = {
...
+		.scratch_addr = 0x54000,
```
apritzel: And I guess we should adjust this as well, to be 4096 bytes after the spl address.

apritzel (changes requested):
Thanks for posting this. I guess we need to adjust some addresses, to be able to later load the same mainline SPL as used on an SD card.

apritzel:
A BootROM dump would indeed be very helpful to confirm the various addresses used, for instance the SD card load address, the stack locations, and where any BROM buffers are. You can try to submit them to https://github.com/smaeul/sunxi-blobs, so that we have them on record for the future

uart0-helloworld-sdboot.c
```diff
@@ -409,7 +413,10 @@ static const struct soc_info *sunxi_detect_soc(void)
 static void clock_init_uart(const struct soc_info *soc)
 {
 	if (soc->flags & FLAG_NEW_CLOCK) {
-		set_wbit(soc->ccu_base + 0x90c,
+		int reg_ofs;
+
+		reg_ofs = (soc->soc_id == 0x1903) ? 0xe00 : 0x90c;
```
apritzel: Please use a new flag for the clock, like FLAG_A733_CLOCK. You can combine this with the other new flag to FLAG_NCAT3 or so.

uart0-helloworld-sdboot.c
```diff
@@ -409,7 +413,10 @@ static const struct soc_info *sunxi_detect_soc(void)
...
+		set_wbit(soc->ccu_base + reg_ofs,
```
apritzel: Please note that each UART has its own BGR register now. So I suggest something like:

    if (soc->flags & FLAG_NCAT3_CLOCK) {
        set_wbit(soc->ccu_base + 0xe000 + uart_index * 4, 0x10001);
    } else if (soc->flags ^ FLAG_NCAT2_CLOCK) {
    ....

uart0-helloworld-sdboot.c
```diff
@@ -446,13 +453,25 @@ static void gpio_init(const struct soc_info *soc)
 		pio_bank_size = 0x30;
 		pio_dat_off = 0x10;
 		pio_pull_off = 0x24;
+	} else if (soc->flags & FLAG_A733_GPIO) {
+		pio_bank_size = 0x80;
```
apritzel, line 460: You also need an offset value, as you figured below, as on the A733, the bank A registers start at 0x80, not 0x0 as before. Either you add a new pio_bank_a_offset variable, and set this 0x80 here, or you adjust pio_base in here, directly.

uart0-helloworld-sdboot.c
```diff
@@ -446,13 +453,25 @@ static void gpio_init(const struct soc_info *soc)
...
 		pio_dat_off = 0x10;
 		pio_pull_off = 0x1c;
 	}
 
+	/*
+	 * For Allwinner SoC A733:
+	 * Adjust pio_base offset by shift size 0x80, as
+	 * PB bank start at 0x100 with the bank size 0x80
+	 */
+	if (soc->soc_id == 0x1903)
```
apritzel: Please no SoC specific hacks, we now have the flags to take of this properly, so that it can be reused later on. I guess this won't be the last SoC using this pinctrl layout. See above for a better solution.

apritzel (changes requested):

soc_info.c
dlan17: > Can you double check that the area at 0x72000 is not used by the BROM? Try to write something to that area, using "sunxi-fel write", then try to read it back (using the "read") command. This should work without using the swap buffers, and this way you can check if the BROM uses that area. Feel free to poke around nearby, it would be good if you could actually find something that the BROM changes, and the confirm that it's NOT between 0x72000 and 0x72400. Another way would be to dump the BootROM, then check the disassembly for addresses used for buffers and stack.

I've done a "write then read back" test with sunxi-fel, and found 0x46E00 - 0x48C00 is the space that we can't write (touched it will corrupt FEL) and should be preserved, so I will use it as swap_buffer

soc_info.c
dlan17: > So this should be 0x47000, to match the load address used by the BROM when loading from SD cards (and other boot media). Making this the same is crucial to be able to use the same SPL binary for FEL and boot media, since we need a fixed link address for the SPL (since we cannot use PIE). And we should double check that it's really 0x47000, and not 0x44000. Again the BROM disassembly would help.

agree, I will use 0x47000

in my test, both 0x44000 and 0x47000 works fine, address 0x44000 will give more space, but downside is that we have to maintain two different value for FEL and booting from SD, so will leave it for now

soc_info.c
dlan17: > And I guess we should adjust this as well, to be 4096 bytes after the spl address.

for spl_addr = 0x47000, then scratch_addr is 0x48000, which < 0x48C00, so it doesn't work..

I would pick 0x71000 if no objection, since it's a temp address for running code

dlan17:
> A BootROM dump would indeed be very helpful to confirm the various addresses used, for instance the SD card load address, the stack locations, and where any BROM buffers are. You can try to submit them to https://github.com/smaeul/sunxi-blobs, so that we have them on record for the future

I'm not familiar with this, I'd appreciate if someone else would do this..

soc_info.c
Icenowy: This (the space that we can't write) is the space that needs to be backed up (buf1 here) instead of the used swap buffer (buf2).

If any higher SRAM is available, then buf2 should be as high as possible.

Icenowy:
At least N-BROM should be easy to dump -- just use `sunxi-fel read` functionality to dump the memory zone of BROM according to the manual.

Icenowy:
Looks like the branch is still quite broken -- `sunxi-fel read` on BROM does not work at all (while `sunxi-fel readl` can read individual words from BROM)

dlan17:
> Looks like the branch is still quite broken -- `sunxi-fel read` on BROM does not work at all (while `sunxi-fel readl` can read individual words from BROM)

$ ./sunxi-fel read 0x17000 0x200 dump.bin
usb_bulk_recv() ERROR -7: Operation timed out

it seems these address space aren't readable? I have no idea why..

Icenowy:
https://github.com/smaeul/sunxi-blobs/pull/5

BROM dumps here (HS-BROM seems to be not available now).

Maybe I should try to do some trick like what @apritzel does on A523, on my Teclast tablet, to retrieve the secure part.

soc_info.c
apritzel: Sure, most addresses will work in FEL, but please note that SD card compatibility is a hard requirement, not a "nice to have": we will only support one image, for FEL, SD card, eMMC and SPI booting.
And the "more space" is quite theoretical, we will probably not need more than say 64KB. For the A523, we are at 37KB at the moment, which leaves us quite some room, even for the probably higher DRAM setup demand of the A733.
So, what is your ``CONFIG_SUNXI_SRAM_ADDRESS`` in U-Boot, and does that work from an SD card? Just use the working version of this address.

dlan17: > So, what is your `CONFIG_SUNXI_SRAM_ADDRESS` in U-Boot, and does that work from an SD card? Just use the working version of this address.

Currently it's CONFIG_SUNXI_SRAM_ADDRESS=0, but while checking the code, it should be CONFIG_SUNXI_SRAM_ADDRESS=0x47000 for case A733?

for FEL mode, it seems doesn't matter of what the CONFIG_SUNXI_SRAM_ADDRESS is, or guess I'm only booting directly from ram.. but SD boot case, due to lacking of DDR initialization, I'm not able to fully test the SD boot procedure.. only got this log

```
U-Boot SPL 2025.10-rc1-00134-gc107a6ec79aa (Aug 27 2025 - 10:53:03 +0000)
DRAM: 1024 MiB
```

then stuck, due to even in SPL stage it will utilize DDR memory space..

apritzel:
Thanks @Icenowy for the BROM dump! That looks like the usual BootROM setup procedure, with all the hotplug checks, and including setting the ACTLR.SMP bit that isn't there anymore on the A55. Oh well ....
Normal stack is at 0xf3400 (towards the end of the NPU SRAM aka "SHARED_SRAM), IRQ stack is at 0x48c00.
> 0x46E00 - 0x48C00 is the space that we can't write

Are you sure it's that large? Typically the IRQ stack usage is pretty small, like at most 1KB (0x400), which would mean we just need to save/restore from 0x48800-0x48c00.
I wonder if we can use the area before the SPL load address for thunk and scratch? So 0x44000 and 0x45000 or so?
I will continue wading through the BROM disassembly, to see if I can find some buffer addresses somewhere, or if it touches more in the lower SRAM area.

dlan17:
> Are you sure it's that large? Typically the IRQ stack usage is pretty small, like at most 1KB (0x400), which would mean we just need to save/restore from 0x48800-0x48c00.

Yes, it's my actual test result, just boot straight into FEL mode (without SD or eMMC), then I did something like this a) "fel write 512bytes" b) "fel read back 512bytes" c) do m5sum to compare

for 0x46e00, it's fail at stage a - fel write

for the large buffer, is the space used by anything else than IRQ stack, like FEL/usb buffer? or maybe simply access forbidden?

> I wonder if we can use the area before the SPL load address for thunk and scratch? So 0x44000 and 0x45000 or so?

I think it's possible, but haven't tested, I will try it some day and will let you know..

dlan17:
> > I wonder if we can use the area before the SPL load address for thunk and scratch? So 0x44000 and 0x45000 or so?
> 
> I think it's possible, but haven't tested, I will try it some day and will let you know..

Yes, it's possible, I can move all buffer used to address before 0x47000, so the diff is something like:


```
+/*
+ * Please note memory from 0x46E00 - 0x46FFF is also used by BROM or FEL
+ * but since we won't touch this range, so no save/restore it
+ */
 sram_swap_buffers a733_sram_swap_buffers[] = {
-       { .buf1 = 0x47000, .buf2 = 0x72400, .size = 0x1C00 },
+       { .buf1 = 0x47000, .buf2 = 0x44000, .size = 0x1C00 },
 };

@@ -756,8 +760,8 @@ soc_info_t soc_info_table[] = {
                .soc_id       = 0x1903, /* Allwinner A733 */
                .name         = "A733",
                .spl_addr     = 0x47000,
-               .scratch_addr = 0x71000,
-               .thunk_addr   = 0x72000, .thunk_size = 0x200,
+               .scratch_addr = 0x46000,
+               .thunk_addr   = 0x45E00, .thunk_size = 0x200,
```

PR #224, opened 2025-08-17
SPI flash rework

apritzel:
The proposed patches to support SPI flash access on more SoCs (#218 and #208) have issues, and are mainly struggling with the slight differences between the SoCs when it comes to clock and pinmux setup. This PR reworks the SPI flash code, mainly by moving SoC specific information into the existing ``soc_info_table``, so the SPI flash code can just read say the GPIO controller base address, without reverting to hard-to-maintain switch/case functions. Also this introduces a flags field, to include the three existing boolean fields, but also add some generation info (like ``H6_STYLE_CLOCKS``), which can now be cleanly attributed to each SoC, again without fragile switch/case constructs.

I started with a separate SoC table in fel-spiflash.c, with just the info we need, but later felt this is just duplicating what we already have. Plus we can use the GPIO and clock base address later, for upcoming features, like MMC support or GPIO access.

Let me know what you think and whether this is a good idea or not.

@iscle @dron0gus please have a look, I think this should simplify the code changes you need to support your SoCs.

### DIFF
--- a/soc_info.c
+++ b/soc_info.c
@@ -334,6 +334,8 @@ soc_info_t soc_info_table[] = {
 		.sram_size    = 48 * 1024,
 		.sid_base     = 0x01C23800,
 		.watchdog     = &wd_a10_compat,
+		.gpio_base    = SUNXI_PIO_BASE,
+		.ccu_base     = AW_CCM_BASE,
 		.flags        = NEEDS_L2EN,
 	},{
 		.soc_id       = 0x1625, /* Allwinner A10s, A13, R8 */
@@ -344,6 +346,8 @@ soc_info_t soc_info_table[] = {
 		.sram_size    = 48 * 1024,
 		.sid_base     = 0x01C23800,
 		.watchdog     = &wd_a10_compat,
+		.gpio_base    = SUNXI_PIO_BASE,
+		.ccu_base     = AW_CCM_BASE,
 		.flags        = NEEDS_L2EN,
 	},{
 		.soc_id       = 0x1651, /* Allwinner A20 */
@@ -355,6 +359,8 @@ soc_info_t soc_info_table[] = {
 		.sid_base     = 0x01C23800,
 		.sid_sections = generic_2k_sid_maps,
 		.watchdog     = &wd_a10_compat,
+		.gpio_base    = SUNXI_PIO_BASE,
+		.ccu_base     = AW_CCM_BASE,
 	},{
 		.soc_id       = 0x1650, /* Allwinner A23 */
 		.name         = "A23",
@@ -365,6 +371,8 @@ soc_info_t soc_info_table[] = {
 		.sid_base     = 0x01C23800,
 		.sid_sections = generic_2k_sid_maps,
 		.watchdog     = &wd_h3_compat,
+		.gpio_base    = SUNXI_PIO_BASE,
+		.ccu_base     = AW_CCM_BASE,
 	},{
 		.soc_id       = 0x1633, /* Allwinner A31 */
 		.name         = "A31",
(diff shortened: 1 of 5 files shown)


dron0gus:
@apritzel thanks for cleaning this mess.
Just tested you branch with my T113 HW. Works fine.
Only two small modifications are needed: https://github.com/dron0gus/sunxi-tools/commit/210586e33b96b0ec6feccde6934cd2a05adb6419 and https://github.com/dron0gus/sunxi-tools/commit/cf1d2adc7eba6b0b03cebe970acd0bf35bec20f0
You can either pick this to your PR or I can create separate PR after this one is merged.

soc_info.c
```diff
@@ -324,6 +324,10 @@ static const sid_section generic_2k_sid_maps[] = {
 	SID_SECTION(NULL, 0, 0),
 };
 
+/* Pack four pin numbers into one uint32_t, using one byte per pin */
+#define SPI_PINS(p1, p2, p3, p4)	\
+	((p1) << 0 | (p2) << 8 | (p3) << 16 | (p4) << 24)
```
dron0gus, line 329: 4 pins are enough for SPI. But T113 also have WP and HOLD signals that are controlled by SPI module.
My test HW have both of these signals connected to flash chip. It works fine without setting correct pinmux for WP and HOLD and leaving it in default state. But I prefer to properly configure these pins.
See https://github.com/dron0gus/sunxi-tools/blob/c6612c5157e78a474b5c2b2129a26b0da5b394a6/fel-spiflash.c#L285
May be extend spi_pins to uint64_t, unconditionally configure 4 first and optionally four other?
```

	/* Setup SPI0 pins muxing */
	for (size_t i = 0; i < 4; i++)
		gpio_set_cfgpin(dev, PC, spi_pin(soc_info, i), soc_info->spi_pinmux);
	for (size_t i = 4; i < 8; i++)
		if (spi_pin(soc_info, i))
			gpio_set_cfgpin(dev, PC, spi_pin(soc_info, i), soc_info->spi_pinmux);
```

dron0gus (approved):

soc_info.c
iscle: Wouldn't a struct with named attributes be better for this? We don't need to differentiate between them in the code (at least for now), but it would be easier to read and know if a pin is used or not.

For example:
```
#define SPI_PIN_NONE -1
// Struct members are signed in order to flag unused ones with -1
typedef struct {
	int8_t cs;
	int8_t clk;
	int8_t mosi;
	int8_t miso;
	int8_t wp;
	int8_t hold;
} spi_pins_t;
```

Just an idea, for me it would be easier to read at first glance. Let me know your thoughts!

apritzel: Well, I don't think the BootROM cares about HOLD and WP, and we try to mimic what the BROM does, for maximum compatibility. Since we don't use QSPI, for our operation both are certainly optional.
And also please keep in mind that sunxi-fel works on a SoC level, not on a board level. Other boards might not connect those pins, or even worse, connect PC7 to the relay that triggers the rocket launch ;-)
So do you really need those pins, for proper operation? Or did you just want to make it neat, by not leaving those pins unconfigured?

dron0gus: @apritzel I agree with your arguments. No boards specific setup should be done here.

DeqingSun:
This PR is awesome.

By adding 

```
		.spi_base     = 0x04025000,
		.spi_pins     = SPI_PINS(0, 1, 2, 3),
		.spi_pinmux   = 4,
```

into Allwinner V853. ```./sunxi-fel spiflash-info ``` can read ID from a W25Q128 on SPI0 of TinyVision V851SE.

apritzel:
@paulkocialkowski can you please have a look at this? There are more SPI support patches coming up, and I would like to clean this up before we gather more hacks in the code base.
I would have a look at your uart0-helloworld rework as well then ;-)

PR #229, opened 2025-12-22
ADD: Basic soc info for H713 chip. Starting point.

Fiooodooor:
ADD: Basic soc info for H713 chip. Starting point.

Chips are being miss recognized and mistakenly wrong values are being used for addresses. This will be added later.

The full credit for this PR must go to the @shift
 owner of https://github.com/shift/sun50iw12p1-research I am only a bystander. Worth noticing that the full repository is result of some kind of vibe coding results and is hard to read and got through.

NOTICE: Please feel free to take ownership of this PR, add changes and make progress. Including merge, rebase of the head etc. Just don't wait for me in case i am not responsive in ~1-2 days

### DIFF
--- a/soc_info.c
+++ b/soc_info.c
@@ -577,6 +577,21 @@ soc_info_t soc_info_table[] = {
 		.rvbar_reg    = 0x08100040,
 		.watchdog     = &wd_h6_compat,
 	},{
+		.soc_id       = 0x1860, /* Allwinner H713 */
+		.name         = "H713",
+		.spl_addr     = 0x20000,
+		.scratch_addr = 0x21000,
+		.thunk_addr   = 0x53a00, .thunk_size = 0x200,
+		.swap_buffers = h616_sram_swap_buffers,
+		.sram_size    = 207 * 1024,
+		.sid_base     = 0x03006000,
+		.sid_offset   = 0x200,
+		.sid_sections = generic_2k_sid_maps,
+		.rvbar_reg    = 0x09010040,
+		.rvbar_reg_alt= 0x08100040,
+		.ver_reg      = 0x03000024,
+		.watchdog     = &wd_h6_compat,
+	},{
 		.soc_id       = 0x1886, /* Allwinner V853 */
 		.name         = "V853",
 		.spl_addr     = 0x20000,


Icenowy:
Where is H713 user manual available? I don't think the PR reviewable w/o UM.

shift:
AFAIK there is no H713 user manual. I even signed up with Allwinner as a developer and couldn't find anything about it. But is is used on a boat load of cheap projectors now running insecure and outdated Android. My repo is linked in the original. I used some spec driven development with a few different agents to get them to reverse engineer the manufacturer supplied image, which luckily had some development headers for the kernel in the initramfs and some other bits which allowed for the RE to be kind of successful. I've had reports from other developers that the U-Boot which my repo can build boots on a few different board revisions of this SoC. Out of the 4 projectors I've had access too each is slightly different. Some have a UART, some don't. Specs of the devices seems to vary wildly.

All of my findings are in the linked repo. You'll need to source the factory images elsewhere due to copyright concerns.

Icenowy:
Well, to be honest, `sp_irq=0x00000000, sp=0x00000000` looks very bogus.
Maybe it's some caching issue?
Could `.icache_fix = true` be tried?

In addition, the SoC info does not match information in the repository too.

Fiooodooor:
As for the Allwinner H713 chip I've got it in cheap projector as @shift mentioned, I bought it as not booting up, out of pure curiosity, for couple of dollars. From the beginning I had "communication" issues when using native Allwinner recommended and available firmware software, but I must admit that changes included in mentioned repository allowed me to make some progress, but not yet managed to patch the firmware and boot it.
If anyone interested I can just send the projector on chosen address as I no longer need it nor have spare time lately.

As a side note, I find Rockchip software way more intuitive.

Fiooodooor:
@icenowy Yes if this is required, as I did not find any relevant docs on this chip.
If I will remember I will send some logs from Linux tools and here is what I could get from Windows 11:

```powershell
PS C:\Users\milos\Downloads\new\xfel-windows-v1.3.5> .\xfel.exe version
WARNING: Not yet support this device ID 0x00186000
ERROR: No FEL device found!
```
And some USB logs:
[hy320_usbtree_driver_log.log](https://github.com/user-attachments/files/24319814/hy320_usbtree_driver_log.log)

Fiooodooor:
More changes with one of waiting PRs added can be seen here: 
https://github.com/linux-sunxi/sunxi-tools/compare/master...Fiooodooor:sunxi-tools:master

apritzel:
@Fiooodooor can you please make a proper pull request? So:
- no merge requests, and no unrelated patches (A733 support)
- fixes first. Just the strncmp change looks like a genuine bugfix (regardless of the buffer change before), please make this a separate patch and make it come first in your PR, or send as a separate PR.
- no fix patches for something you introduced. So the H713 support should be *one* patch, not one patch and a fix.

I don't understand what that "H616 variant" in the name should mean.
And I am confused about the SPL address. If the chip has SRAM A1, at 0x20000, we should load the SPL in there, as this is probably what the BootROM does for SD card boots. U-Boot's SPL is linked at a fixed address, so we need that to be the same between SD card, eMMC, NOR SPI flash (dictated by the BROM), and FEL.
The sentence "boot0.bin load SPL at some address" doesn't make much sense, because boot0 is the Allwinner name for the SPL. What's important is where boot0.bin has been loaded to, and we should use that address. It is possible that this is in some large SRAM at 1MB, as the R329, with a close-by SoC-ID, uses that address, but we would need to confirm this.
A BootROM dump and/or disassembly would help here.

Fiooodooor:
@apritzel I would of course love to make a proper, full, pull request from the mentioned fork changes (now updated to exclude out of scope changes)
https://github.com/linux-sunxi/sunxi-tools/compare/master...Fiooodooor:sunxi-tools:master 

but the current one is introducing only basic info for the chip id to work: 
image
Please feel free to add any required changes to my PR as I am doing this only in free time.
@shift any updates from your side regarding the topic? Can you check the link i've shared and give me your thoughts :)

apritzel:
@Fiooodooor this "basic" info is mostly all we need for a new SoC, but we need to make sure it's correct. I don't have that device, so we would need to lean on someone who does to verify that. So for instance what is the spl_addr you propose? 0x20000 or 0x100000? And the screenshot looks like a copy of H616, including the H616 RVBAR hack we needed, but I doubt the H713 needs. This could all be checked better with a BROM dump. Does FEL somehow work for you, with those changes? Can you please try to dump the BROM then, with:

    $ ./sunxi-fel read 0 32768 brom.bin

Maybe we need more than 32K, please try bigger values as well.

And reading this AI vibe coding ... nonsense left me seriously confused.
How do you enter FEL on this projector? By connecting a USB-A <-> USB-A cable, without powering on the device, as suggested by some reports? This might mean the device is powered through the voltage injected via the USB-A port, from your host, which might not be enough to provide stable power? So regular boot fails, and the BROM enters FEL, but even for FEL it's not enough power? And thus we see all kinds of strange things, including reboots?
But is all speculation and shooting in the dark, unless we get reliable reports of what's going on exactly.

Also the strcmp change looks weird, after having a second look. "AWUS" should be followed by 0x00 bytes, the wiki reports this being a tag. In my experiments this got echoed back from the "AWUC" request sent earlier. So by us putting 0x00's in there we should get them back, and the strcmp should work. So can you dump what was actually sent back, especially behind the "AWUS", when the original code triggered the assert?

Oh, and also:
> doing this only in free time.

this is true for almost everyone messing around with Allwinner devices, so welcome to the club ;-)

jefflessard:
Hi @apritzel, I'd like your review for this H713 alternative entry, with each value explicitly sourced. I'm new to sunxi platform and sunxi-fel, so I am not mastering everything here.

According to my analysis, the H713 (`sun50iw12p1`) SRAM layout is completely different from H616: `spl_addr` is `0x104000`, not `0x20000`, so the existing PR's entry is a straight copy-paste from H616 and is incorrect.


Here's the derived entry with rationale for each value:

```c
sram_swap_buffers h713_sram_swap_buffers[] = {
    /*
     * BROM FEL scratchpad sits at 0x121500 (reported directly by
     * "sunxi-fel -v version"). buf1 starts at 0x121000 to cover it
     * with a 4KB window. buf2 at 0x10c000 is the safe middle zone:
     * above SPL's 32KB init RAM (0x104000 + CONFIG_SYS_INIT_RAM_SIZE
     * 0x8000 = 0x10c000 = CONFIG_FES1_RUN_ADDR) and below the BROM
     * scratchpad.
     */
    { .buf1 = 0x121000, .buf2 = 0x10c000, .size = 0x1000 },
    { .size = 0 }
};

/*
 * Watchdog base 0x02051000 confirmed via ADB device-tree inspection
 * (watchdog@2051000, compatible = "allwinner,sun50i-wdt").
 * reg_mode = base + 0x08: derived from Creality BSP
  * https://github.com/CrealityTech/sonic_pad_os/blob/main/lichee/brandy-2.0/u-boot-2018/drivers/watchdog/sunxi_wdt.c struct layout
 * (ctl @ +0x00, cfg @ +0x04, mode @ +0x08). reg_mode_value = 1
 * (WDT_MODE_EN): the write-protect key (0x16aa) is only applied to
 * wdt->ctl in this driver, not wdt->mode, so no key needed here.
 */
const watchdog_info wd_h713_compat = {
    .reg_mode       = 0x02051008,
    .reg_mode_value = 1,
};

{
    .soc_id = 0x1860,               /* confirmed via ADB: sram_ctrl soc_ver/ver_a = 0x18600000,
                                     * and by sunxi-fel -v version output */
    .name   = "H713",
    /*
     * spl_addr: CONFIG_BOOT0_RUN_ADDR = CONFIG_SYS_SRAM_BASE = 0x104000,
     * from sun50iw12p1.h in Creality BSP
     * https://github.com/CrealityTech/sonic_pad_os/blob/main/lichee/brandy-2.0/spl/include/configs/sun50iw12p1.h
     * struct sunxi_wdog: {ctl@+0x00, cfg@+0x04, mode@+0x08} consistent
     * with wd_a523_compat (base 0x02050000, reg_mode = 0x02050008 = base+0x08).
     * sram_size: CONFIG_SYS_SRAM_SIZE = 0x020000 (128 KiB), same source.
     */
    .spl_addr     = 0x104000,
    /*
     * scratch_addr: taken directly from BROM FEL report ("scratchpad=00121500"),
     * rather than the usual spl_addr + 0x1000 = 0x105000
     * "sunxi-fel -v version" → scratchpad=00121500
     */
    .scratch_addr = 0x121500,
    /*
     * thunk_addr: SRAM top (0x104000 + 0x20000 = 0x124000) minus thunk_size.
     */
    .thunk_addr   = 0x123e00, .thunk_size = 0x200,
    .swap_buffers = h713_sram_swap_buffers,
    .sram_size    = 128 * 1024,
    /*
     * sid_base: confirmed via ADB device-tree (sid@3006000, reg = 0x03006000,
     * size = 0x1000). sid_offset 0x200 follows the universal post-H6 pattern.
     * Full SID section map not documented in BSP DT; using generic placeholder.
     */
    .sid_base     = 0x03006000,
    .sid_offset   = 0x200,
    /*
     * rvbar_reg: assumes the H713 doesn't need the H616 RVBAR hack, using the
     * single-register pattern from, A133/R329 (0x08100040) with no rvbar_reg_alt.
     */
    .rvbar_reg    = 0x08100040,
    /*
     * ver_reg: sram_ctrl base 0x03000000 + soc_ver/offset 0x24 (confirmed
     * via ADB: sram_ctrl@3000000/soc_ver/offset = 0x00000024).
     */
    .ver_reg      = 0x03000024,
    .watchdog     = &wd_h713_compat,
},
```

The entry has been validated on a live H713 device reporting androidboot.hardware=sun50iw12p1: `sunxi-fel -v version` identifies the SoC correctly and FEL reads complete without hangs, confirming `swap_buffers` preserves the BROM stacks correctly. The one field I'm least confident about is `rvbar_reg`, I'd be happy to defer to your judgment there since it can't be verified without actually exercising multi-core bring-up through FEL.

Here are the tests I've run so far with this alternative entry:

```sh
$ sunxi-fel -v read 0x104000 4096 sram_low.bin && xxd sram_low.bin | head
00000000: 0300 0000 4350 5573 1803 0000 0300 0000 ....CPUs........
00000010: fb7b 7b00 0100 0000 f410 0000 0000 0004 .{{.............
00000020: 701c 0000 4000 0000 1800 0000 0000 0000 p...@...........
00000030: 9521 4a00 9031 4202 61b0 0800 9678 78b4 .!J..1B.a....xx.
00000040: 0000 0000 4848 4848 4800 0000 1e12 2016 ....HHHHH..... .
00000050: 0000 0000 0000 0000 0000 0000 0000 3444 ..............4D
00000060: 6666 0000 0361 01b4 0000 0000 0000 0000 ff...a..........
00000070: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000080: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000090: 0000 0000 0000 0000 0000 0000 0000 0000 ................
```

```sh
$ sunxi-fel -v read 0x121000 4096 sram_top.bin && xxd sram_top.bin | head
00000000: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000010: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000020: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000030: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000040: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000050: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000060: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000070: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000080: 0000 0000 0000 0000 0000 0000 0000 0000 ................
00000090: 0000 0000 0000 0000 0000 0000 0000 0000 ................
```

```sh
$ sunxi-fel -v read 0 32768 brom.bin
usb_bulk_recv() ERROR -7: Operation timed out

$ sunxi-fel -v read 0x00010000 65536 brom.bin # try H6/H616 BROM location
usb_bulk_recv() ERROR -7: Operation timed out

$ sunxi-fel -v read 0xffff0000 65536 brom.bin # ARM high-vec fallback
usb_bulk_recv() ERROR -7: Operation timed out
```

apritzel:
Hi @jefflessard, thanks for the info, a few things to mention:
- `rvbar_reg` is needed for the 64-bit switch, sunxi-fel doesn't care about secondary cores
- `sunxi-fel read` is mapping directly to a FEL USB protocol primitive, there is no code upload or swap_buffers used. It would work once you have some entry with the  SoC ID matching, so doesn't prove that the other values are correct.
- `scratchpad` and `scratch_addr` are two completely different things. The former is just one address that the BROM reports, and we print this, purely for informational purposes. I think it means the BROM code uses this address for buffers or BSS or something, so we actually should not touch it. `scratch_addr` on the other hand is some free SRAM region, purely designated by us, that we can use to upload small code snippets. It should *not* be used by the BROM in any way.
-  The reason for swap_buffer is to save and restore the buffers and stacks used by the BROM that are in the area we load the SPL to. If something is far away from that, we don't need to swap it. So as this BROM scratchpad is almost 120K away from the SPL load address, I wouldn't worry about it - just leave it alone (limit SRAM size). 
- The `.buf2` address should be *high*, far away from the low SRAM addresses used by any code like the SPL. 32K is far too limiting for modern SoCs.
- So the swap_buffers typically need to cover the IRQ stack, which historically lives close to the beginning of the SRAM, like a few KBytes in. I think the `spl` command dumps the value of the SP_IRQ register? We would definitely need to save that IRQ stack. Some BROMs use buffers in low SRAM as well, this could be verified by repeatedly dumping the SRAM and looking for changes. Or writing known values to the SRAM and trying to read them back, to compare the content.
- I would not give too much on similarities with previous SoCs, Allwinner changes things seemingly randomly, when they seem fit. Especially I think the H616 is not particularly close: the BSP mention NCAT2, and also the SoC ID puts that chip more into the vicinity of the R329 and R528/D1/T113s3. R329 also has high SRAM, so that might be a better candidate for copy&paste.

There are more things, I think you gather that's not a trivial task.
Do you have a serial console connected? IIRC there is no SD card on this device, so we cannot boot from there to verify things?
I have some diagnostic code that dumps registers into SRAM, where we can read them back, so we don't need serial. I will try to dig this out and send it over, to learn about the BROM specifics here, and maybe eventually even being able to dump the BROM. I think FEL blocks direct access deliberately, but we can upload code that copies the BROM to SRAM, then read this back afterwards. If you feel comfortable with bare metal Arm32 assembly, give it a go, otherwise I see what existing code I can find here for that.

jefflessard:
Hi @apritzel, thanks for the detailed feedback, it was very helpful. I've reworked the entry based on your corrections. Here's the updated state with methodology for each value.

## Corrected Entry

```c
sram_swap_buffers h713_sram_swap_buffers[] = {
    /* sp_irq=0x00105400 confirmed via "sunxi-fel spl" on clean FEL session.
     * SRAM diff (multiple dumps across FEL operations) independently confirms
     * BROM modifies 0x10539C-0x1053FC during FEL — consistent with IRQ stack
     * just below 0x105400.
     * buf2 at 0x11E000: high SRAM, well below main stack (sp=0x120300).
     */
    { .buf1 = 0x105000, .buf2 = 0x11e000, .size = 0x0400 },
    { .size = 0 }
};

const watchdog_info wd_h713_compat = {
    .reg_mode       = 0x02051008,
    .reg_mode_value = 1,
};

{
    .soc_id       = 0x1860,
    .name         = "H713",
    .spl_addr     = 0x104000,
    .scratch_addr = 0x106000,
    .thunk_addr   = 0x120e00, .thunk_size = 0x200,
    .swap_buffers = h713_sram_swap_buffers,
    .sram_size    = 116 * 1024,
    .sid_base     = 0x03006000,
    .sid_offset   = 0x200,
    .icache_fix   = true,
    .rvbar_reg    = 0x08100040,
    .ver_reg      = 0x03000024,
    .watchdog     = &wd_h713_compat,
},
```

## What changed from my earlier proposal and why

**`scratch_addr`: `0x121500` → `0x106000`**

You were right that the BROM-reported scratchpad and `scratch_addr` are unrelated. I set `scratch_addr = 0x106000` because code execution at `0x105000` fails (it overlaps the IRQ stack at `sp_irq = 0x105400`), while `0x106000` works reliably. Tested by uploading and executing small ARM snippets that write known values to a target address.

**`swap_buffers`: completely reworked**

Following your guidance, I determined the IRQ stack location empirically using two methods:

1. `sunxi-fel spl fes1.fex` reports `sp_irq=0x00105400, sp=0x00120300` (with `icache_fix`, see below).
2. SRAM pattern-diff: dumped the full SRAM, ran FEL operations (version only, no scratch_addr usage), dumped again, diffed. The BROM modifies exactly these locations between operations:
   - `0x10539C`, `0x1053C8-0x1053CA`, `0x1053DC`, `0x1053FC` (IRQ stack area)
   - `0x120248`, `0x120314` (main stack area)

   This was consistent across 5 consecutive dump pairs.

So `buf1 = 0x105000` with `size = 0x400` covers the IRQ stack, and `buf2 = 0x11E000` is high SRAM (104 KiB from spl_addr), well below the main stack.

**`sram_size`: `128 * 1024` → `116 * 1024`**

Per your advice to "limit SRAM size" to exclude the BROM scratchpad area. The BROM reports its scratchpad at `0x121500`, so I cap at `0x121000` (giving `0x121000 - 0x104000 = 0x1D000 = 116 KiB`). `thunk_addr` adjusted accordingly to `0x120E00`.

**`icache_fix = true`** (new)

This was a significant finding. Without `icache_fix`, the `aw_get_stackinfo` thunk returns garbage, it reads back stale data from a previous operation at `scratch_addr` instead of the freshly uploaded code's results. With `icache_fix = true`, `sp_irq` and `sp` come back as valid SRAM addresses that match the SRAM diff results. I also observed that code execution at a given address works on first use but fails on subsequent uploads of different code to the same address within the same FEL session, classic icache staleness.

**`watchdog`: `0x02051008`**

Base `0x02051000` confirmed via ADB device-tree (`watchdog@2051000`, compatible `allwinner,sun50i-wdt`). Mode register offset `+0x08` from BSP `sunxi_wdt.c` struct layout (`ctl@+0x00, cfg@+0x04, mode@+0x08`), cross-referenced against the A523 entry which uses the same `base + 0x08` pattern. Not destructively tested.

## What I don't have

- **No serial console**: There are TRG pads on the PCB (HY300 Pro+) which I guess are the UART but I could not get any output. Trying to see what's happening on the line with an oscilloscope didn't help.
- **No SD card slot**: eMMC only.
- **`rvbar_reg`**: using `0x08100040` per your earlier comment. Cannot verify without actually executing the 64-bit switch.
- **`sunxi-fel spl` upload fails**: stack pointers are read successfully, but the subsequent SPL data upload times out during `usb_bulk_send`. This happens after the swap_buffers save and during the actual data write to the SPL area. Basic FEL writes work fine (I've written 638 KB to DRAM after fes1 init without issues), so this seems specific to the SPL upload sequence. Any ideas?

## Device details

- HY300 Pro+ projector, Allwinner H713, 1 GB DDR3, eMMC only
- Android 11, rooted, ADB available
- FEL mode accessible via physical button
- Kernel: `5.4.99-00049-g34f0974adef4` (vendor BSP, 32-bit userspace)
- Bootargs report `androidboot.hardware=sun50iw12p1`

Your offer of diagnostic code for register/BROM dumping would be very welcome. I'm comfortable running bare-metal ARM32 payloads via `sunxi-fel exe` (that's how I validated `scratch_addr` and `icache_fix`). Being able to dump the BROM would be particularly useful since FEL blocks direct reads to the BROM address range (reads at `0x0`, `0x10000`, `0xFFFF0000` all time out).

Issue #217, opened 2025-02-23
A523/A527 SPI Flash support?


RoganDawes:
Hi folks,

Is there anyone looking at implementing SPI flash support for the A523/A527, and possibly T527 chip? I have just received my pre-order of the Radxa A5E, and am trying to write their NOR image to the SPI flash. My distro didn't recognise the chip, so I cloned the repo, built and tried spiflash-info, only to be told that it is not yet supported.

Thanks!

Rogan

apritzel:
The manual suggests that the SPI controller in the A523/A527/T527 (all the same chip) is compatible to the previous one, however it doesn't work. Poking around in the BSP reveals that apparently the FIFO status register at 0x1c no longer reports the FIFO level, instead this is now done with a new register at offset 0x400.
So this would require some changes in the code, on top of the expected SoC adaptations (pinmux, clocks). I have a (non-working) patch for the latter, but didn't try the new FIFO register yet.

PopCat19:
Just confirming this issue on another A523 device, an unbranded "Eksvefot A10" tablet I got from Amazon:
```
~ 
❯ sunxi-fel version
AWUSBFEX soc=00001890(A523) 00000001 ver=0001 44 08 scratchpad=00061500 00000000 00000000

~ 
❯ sunxi-fel spiflash-read 8192 32768 u-boot-spl.bin
SPI support not implemented yet for 1890 (A523)!
```

I'm hitting the same `SPI support not implemented yet for 1890 (A523)!` error on `spiflash-read` in attempts to get a firmware dump.

While I can't code the FIFO register fix myself, I'm open to build and test any experimental patches if you need someone to validate on this hardware.

ddean-blip:
did this ever get any further im trying to back up a device now with the A527 chipset and running into issues .

/a527_dump$ sudo sunxi-fel spl u-boot-sunxi-with-spl.bin
Warning: no 'soc_sram_info' data for your SoC (id=1890)
SPL: eGON header is not found
/a527_dump$ sudo sunxi-fel version
Warning: no 'soc_sram_info' data for your SoC (id=1890)
AWUSBFEX soc=00001890(unknown) 00000001 ver=0001 44 08 scratchpad=00001500 00000000 00000000
/a527_dump$ sunxi-fel spiflash-read xxxx xxxx u-boot-spl.bin
ERROR: You don't have permission to access Allwinner USB FEL device

apritzel:
[...]

PR #232, opened 2026-04-05
feat: A523 SPI support

arminfelder:
adds SPI flash support for A523/A527/T527 SoCs 

tested/developed against OrangePi 4A, using u-boot mainline and BL31 from [github.com/jernejsk/arm-trusted-firmware](https://github.com/jernejsk/arm-trusted-firmware/tree/a523-v4)

The information was extracted from the A523 datasheet as well as the Tina SDK(using an LLM), and tested against the real chip(T527)

relates to #217

### DIFF
--- a/fel-spiflash.c
+++ b/fel-spiflash.c
@@ -87,6 +87,10 @@ void fel_writel(feldev_handle *dev, uint32_t addr, uint32_t val);
 #define H6_CCM_SPI_BGR              (0x03001000 + 0x96C)
 #define H6_CCM_SPI0_GATE_RESET      (1 << 0 | 1 << 16)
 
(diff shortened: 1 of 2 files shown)


RoganDawes:
Can confirm that it works. Was able to write u-boot to the SPI of my Radxa A5E. Thank you!

apritzel:
Hi, thanks for pushing this. But while this looks like a pragmatic solution to the problem, it's becoming messier with every SoC, piling up exceptions and special cases.
Please have a look at https://github.com/linux-sunxi/sunxi-tools/pull/224 for my proposed cleanup. This would mean adding A523 support would be much easier. Can you change your code to work on top of that?

arminfelder:
I guess I wait for the rework to be merged, and then rebase

PR #233, opened 2026-04-13
fel: add progress bar support for uboot command

gworkman:
The sunxi-fel uboot (and spl) commands lacked the ability to show a progress bar when uploading the payload, unlike the write command.

Support the -p / --progress flag by passing a progress callback through aw_fel_process_spl_and_uboot to aw_fel_write_uboot_image, which then sets up the progress indicator before calling aw_write_buffer.

Result:
```
$ sunxi-fel -p uboot path/to/u-boot-loader.bin
100% [================================================]   535 kB,  450.4 kB/s
```

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -904,7 +904,8 @@ uint32_t aw_fel_write_and_execute_spl(feldev_handle *dev, uint8_t *buf, size_t l
  * U-Boot entry point (offset) and size values.
  */
 static void aw_fel_write_uboot_image(feldev_handle *dev, uint8_t *buf,
-				     size_t len, const char *dt_name)
+				     size_t len, const char *dt_name,
+				     progress_cb_t progress)
 {
 	if (len <= HEADER_SIZE)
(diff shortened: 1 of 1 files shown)


apritzel:
Is there any particular need for that? I see that's somewhat inconsistent with the other transfer commands, but as your screenshot above shows, U-Boot typically uploads within one second, so I am not sure we really need a progress bar? The moment you look at it, it's typically finished already.

gworkman:
Hi @apritzel, sorry the delay in getting back to you! This got buried in my GitHub notifications.

Yeah, I can understand the case for not needing a progress bar since it is very fast. My use case is this: I am using my fork right now with some CLI tooling to implement a GUI for a custom flashing app for my open-source hardware project, so users can flash their own firmware. Currently the flashing app parses the stdout from the CLI and show a different progress bar in the GUI so the user knows that something is happening (even if that delay is only 1-2 seconds)

If this isn't a relevant feature for the project and you wish to close the PR, totally understand :)

PR #237, opened 2026-05-18
sunxi-fel: find libfdt with pkg-config

jameshilliard:
sunxi-fel uses libfdt to parse FIT images, but the Makefile only passed -lfdt when linking. That works when libfdt headers and libraries are installed in default compiler search paths, but fails when a package installs them elsewhere.

Use pkg-config to obtain the libfdt compiler and linker flags, matching the existing libusb and zlib handling.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,6 +134,16 @@ ZLIB = zlib
 ZLIB_CFLAGS ?= `$(PKG_CONFIG) --cflags $(ZLIB)`
 ZLIB_LIBS ?= `$(PKG_CONFIG) --libs $(ZLIB)`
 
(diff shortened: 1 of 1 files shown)


apritzel (approved):
Thanks, that looks good now, and I like it much better than the previous MacOS/homebrew specific solution.
Briefly tested without any libfdt.pc file, and with one installed.

paulkocialkowski (approved):
Looks good to me as well, thanks for reworking this!

paulkocialkowski:
Maybe we can wait a few days in case someone else wants to weigh-in and we can merge it afterwards.

PR #238, opened 2026-05-18
Makefile: prefer mkimage for sunxi images

jameshilliard:
U-Boot's mkimage supports generating Allwinner eGON boot images with the sunxi_egon image type. Prefer that when available, since mkimage is maintained and commonly packaged with U-Boot tools.

Keep mksunxiboot as a fallback for setups with older mkimage versions or without mkimage installed.


apritzel (approved):
Thanks, looks good, and mkimage is indeed the more reliable tool moving forward. Tested by misspelling mkimage and sunxi_egon, to test the fallback paths.

PR #239, opened 2026-05-18
Makefile: avoid broad wildcards in clean

jameshilliard:
The clean target currently removes every *.bin, *.elf, *.sunxi, *.nm and *.orig file in the source directory. That can delete unrelated files placed in the tree, such as downloaded binaries or local test artifacts.

Replace those broad wildcards with an explicit list of files generated by this Makefile. Also include sunxi-script_extractor, which has an explicit build rule but was not previously listed in the clean target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,10 @@ MISC_TOOLS = phoenix_info sunxi-nand-image-builder
 # ARM binaries and images
 # Note: To use this target, set/adjust CROSS_COMPILE and MKSUNXIBOOT if needed
 BINFILES = jtag-loop.sunxi fel-sdboot.sunxi uart0-helloworld-sdboot.sunxi
+BINFILE_ELFS = $(BINFILES:.sunxi=.elf)
+BINFILE_BINS = $(BINFILES:.sunxi=.bin)
+BOOT_HEAD_ELFS = boot_head_sun3i.elf boot_head_sun4i.elf boot_head_sun5i.elf
+CLEANFILES = $(BINFILES) $(BINFILE_ELFS) $(BINFILE_BINS) $(BOOT_HEAD_ELFS)
 
 MKSUNXIBOOT ?= mksunxiboot
 PATH_DIRS := $(shell echo $$PATH | sed -e 's/:/ /g')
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -106,8 +113,7 @@ install-misc: $(MISC_TOOLS)
 
 clean:
 	make -C tests/ clean
-	@rm -vf $(TOOLS) $(FEXC_LINKS) $(TARGET_TOOLS) $(MISC_TOOLS)
```
apritzel, line 109: can you please keep this line? That separates the executables from the intermediary files and makes the diff smaller

Makefile
```diff
@@ -53,6 +53,13 @@ MISC_TOOLS = phoenix_info sunxi-nand-image-builder
 # ARM binaries and images
 # Note: To use this target, set/adjust CROSS_COMPILE and MKSUNXIBOOT if needed
 BINFILES = jtag-loop.sunxi fel-sdboot.sunxi uart0-helloworld-sdboot.sunxi
+BINFILE_ELFS = $(BINFILES:.sunxi=.elf)
+BINFILE_BINS = $(BINFILES:.sunxi=.bin)
+BOOT_HEAD_ELFS = boot_head_sun3i.elf boot_head_sun4i.elf boot_head_sun5i.elf
+NAND_PART_OBJS = nand-part-main.o nand-part-a10.o nand-part-a20.o
```
apritzel: I think it's common to remove all .o files. I see your point in limiting the .sunxi, and maybe .bin and .elf files to be removed (I was hit by this myself before), but there shouldn't really be foreign object files in this directory. So I'd drop this NAND_PART_OBJS line and use *.o, ideally directly below.
Explicitly listing the generated .o files here sounds like a maintenance burden going forward.

Makefile
```diff
@@ -106,8 +113,7 @@ install-misc: $(MISC_TOOLS)
 
 clean:
 	make -C tests/ clean
-	@rm -vf $(TOOLS) $(FEXC_LINKS) $(TARGET_TOOLS) $(MISC_TOOLS)
-	@rm -vf version.h *.o *.elf *.sunxi *.bin *.nm *.orig
```
apritzel, line 110: Can you please keep version.h and *.o named here? Then just use the new variable to cover the .elf, .sunxi and .bin files.
It's fine to drop .nm and .orig, but please mention that in the commit message.

apritzel (changes requested):
Thanks, that's a useful change - as I have been bitten by this myself before.
Just some smaller comments inside.

Makefile
jameshilliard: Restored this line.

Makefile
jameshilliard: Restored the wildcard for .o files.

Makefile
jameshilliard: Applied as suggested.

apritzel (approved):

PR #240, opened 2026-05-24
Makefile: avoid GNU install options

jameshilliard:
The install-tools target installs the sunxi-fel man page with "install -D -m0644 -t", which is supported by GNU install but not by BSD install. This breaks "make install" on systems such as macOS.

Create the man page directory explicitly and install the file to its full destination path instead. This matches the existing directory creation used for the binary install path and works with both GNU and BSD install.


apritzel (approved):
Yes, looks good. `-D` indeed seems non-standard, FreeBSD has it, for instance, but with a different semantic. So using the more elaborate version is better.

paulkocialkowski (approved):

PR #234, opened 2026-05-14
Optimize FEL USB performance

jameshilliard:
Implementation details: [README.FEL-SPEED.md ](https://gist.github.com/jameshilliard/bc551c260679e53740b63276299b8fe6)

With optimizations:
```c
sunxi-fel -v -p uboot u-boot-sunxi-with-spl.bin write 0x49000000 fit.itb
Tracking FEL device by SID 32c05000:0c004808:01075b8a:2c9a1f11
Switching USB to high-speed mode, setting FEL bulk endpoints to 512-byte packets: 0x05100040: 0x00029940 -> 0x00029860
Selecting FEL device 000:002 by SID
Applying SMC workaround via secure-SVC return thunk...  done.
TOC0: wrapped SPL item 49111 bytes from 0x840, load 0x00020060
TOC0: payload starts at 0xe000
Stack pointers: sp_irq=0x00021400, sp=0x00053FFC
MMU is not enabled by BROM
=> Executing the SPL... done.
loading image "ARM Trusted Firmware" (41065 bytes) to 0x40000000
loading image "U-Boot" (917496 bytes) to 0x4f000000
loading DTB "sun50i-h616-whatsminer" (48408 bytes)
Patching BROM RX FIFO copy to use DMA.
100% [================================================] 41082 kB, 30185.6 kB/s 
Starting U-Boot (0x40000000).
Store entry point 0x40000000 to RVBAR 0x09010040, and request warm reset with RMR mode 3... done.
```

Without optimizations:
```c
sunxi-fel --no-high-speed -v -p uboot u-boot-sunxi-with-spl.bin write 0x49000000 fit.itb
Applying SMC workaround via secure-SVC return thunk...  done.
TOC0: wrapped SPL item 49111 bytes from 0x840, load 0x00020060
TOC0: payload starts at 0xe000
Stack pointers: sp_irq=0x00021400, sp=0x00053FFC
MMU is not enabled by BROM
=> Executing the SPL... done.
loading image "ARM Trusted Firmware" (41065 bytes) to 0x40000000
loading image "U-Boot" (917496 bytes) to 0x4f000000
loading DTB "sun50i-h616-whatsminer" (48408 bytes)
100% [================================================] 41082 kB,  320.0 kB/s 
Starting U-Boot (0x40000000).
Store entry point 0x40000000 to RVBAR 0x09010040, and request warm reset with RMR mode 3... done.
```

PR #241, opened 2026-05-28
fel: add FEL protocol and soc_info_t struct documentation

apritzel:
Add a markdown document describing the basics of the FEL protocol, to give some idea about the expectations and limitations of the protocol.

Also add a description for each member of the soc_info_t struct, and explain how their values should be determined for a new SoC.

### DIFF
--- a/README.FEL.md
+++ b/README.FEL.md
@@ -1,0 +1,175 @@
+# Allwinner FEL operation
+
+The FEL protocol is a proprietary USB protocol developed and solely used by
+Allwinner. The device side is implemented in the Boot-ROM of all Allwinner SoCs.
+The purpose of this protocol is to give easy debug and bootstrapping access
+to the SoC's operation, allowing to upload and execute code without any other
+media connected. This eventually allows to boot operating systems solely via
+the USB interface, or, with the right payload code, to access peripherals on
+the board.
+
+Upon leaving the initial reset state, the primary core starts executing
+Boot-ROM code. One of the first checks is for the status of the "FEL button"
+pin: if this is connected to GND, the Boot-ROM starts executing FEL code. The
+same is true when all of the other media boot methods fail, or when user code
+later jumps to a certain Boot-ROM address (typically offset 0x20).
+
+The Boot-ROM FEL code sets up the MUSB USB0 controller in peripheral mode, and
+waits for connections from a host. The protocol uses four commands:
+- `AW_FEL_VERSION`: returns very basic protocol and SoC information. This mostly
+  just contains a signature and a protocol version number, to confirm that this
+  is really a device using the FEL protocol, and the 16-bit SoC-ID, to identify
+  the SoC model.
+- `AW_FEL_1_WRITE`: writes data into the specified memory area of the SoC.
+  Theoretically any memory location can be used, including MMIO regions, but the
+  accesses are done using byte-wide accessors, so are incompatible with most
+  devices' requirements for 32-bit wide accesses. So in practice those writes
+  must be to either SRAM or DRAM.
+- `AW_FEL_1_READ`: reads data from the specified memory area of the SoC. The
+  same byte-wide accesses as for writes are used, so the same limitations
+  apply: only read from SRAM or DRAM.
+- `AW_FEL_1_EXEC`: branches to the specified memory address to execute code.
+  The Boot-ROM runs in 32-bit mode, so at least the initial instructions
+  must be in AArch32 ARM mode, even on 64-bit SoCs.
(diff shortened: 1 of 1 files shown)


README.FEL.md
```diff
@@ -0,0 +1,175 @@
...
+
+      AWUSBFEX soc=00001855(unknown) 00000001 ver=0001 44 08 scratchpad=00007e00 00000000 00000000
+                       ^^^^
+
+  Sometimes this value can also be found in BSP source code, or read from the
+  SID, with a special sequence.
+- `name`: A string with the short name of the SoC, like `"A133"` or `"T527"`.
+- `spl_addr`: This is the address where the BootROM(BROM) will load an eGON
+  wrapped payload to. Often this is the start of the first SRAM block (SRAM A1),
+  though newer SoCs tend to deviate here. This must be the same value where the
```
dlan17, line 76: for "must be the same value", it would be great to emphasis this choice is taken by sunxi-tools's policy for easing maintenance, not restricted by lowlevel BROM protocol, which I mean it's totally ok to download code to any valid address and execute..

README.FEL.md
```diff
@@ -0,0 +1,175 @@
...
+  BROM will load boot0/SPL from SD card, eMMC or SPI flash, to allow the same
+  SPL binary to work with any load method.
+  The SRAM address(es) should be mentioned in the memory map of the manual, but
+  it's best to confirm this by loading a small code snippet from SD card and
+  printing the load address.
```
dlan17, line 81: is there any public repo available for that small code which able to print the address? or docs recorded for this
(I remember you shared that when I add fel support for A733..)

dlan17 (commented):
This doc is really excellent for people to get overview of soc_info structure, so I realize it's not really necessary to get into too many details which takes extra more efforts, or even we can just leave to improve them later.

README.FEL.md
ssvb: It's not OK to load the U-Boot's SPL to a wrong address because it's not position independent.

README.FEL.md
```diff
@@ -0,0 +1,175 @@
...
+  some areas near the beginning and some areas towards the end of the SRAM.
+  A 512-byte area below the lowest address of the end region is a good choice.
+- `thunk_size`: The available space at `thunk_addr`. 512 bytes are recommended,
+  but a smaller region might be sufficient, if space is really tight.
+- `needs_l2en`: Boolean flag to determine whether payload code must enable
+  the L2 cache. This is only needed for the two oldest SoCs, and can be left
+  out (set to `false`) for every other SoC.
+- `mmu_tt_addr`: Address of some memory area to hold the page tables during
+  the FEL operation. Only needed on older SoCs that enable and require the
+  MMU during FEL operation. Can be left out (set to 0) for newer SoCs.
```
ssvb, line 107: It's not needed on old SoCs either. But it provided significantly faster data transfer speed. See the detailed explanations in the following commit messages:
https://github.com/linux-sunxi/sunxi-tools/commit/e4b3da2b17ee9d7c5cab9b80e708b3a309fc4c96, https://github.com/linux-sunxi/sunxi-tools/commit/a275b246a4c8af44162973ffdc1aaa9da4a074f9, https://github.com/linux-sunxi/sunxi-tools/commit/c32eeb88ff6d90f7b0c4cc64ee0719b2f944aa80. And also the table at https://linux-sunxi.org/FEL/USBBoot#SoC_support_status
The difference between ~180KB/s and ~900KB/s, observed on H3 was gigantic from the practical point of view, considering the typical sizes of the linux kernel images.

jameshilliard: If you have a H3 you should maybe see if [DMA](https://github.com/linux-sunxi/sunxi-tools/pull/234) works on it. I get nearly a 100x speed increase with that on H616 and ported the optimizations based on BROM dump analysis.

README.FEL.md
apritzel: I agree with @ssvb: it's essential to use the same address. Loading mainline U-Boot is the main feature of this tool, so we have to live with the restrictions. If you want to download and execute arbitrary code, you can always do so via the FEL primitives (write & exec), and then you can pick you own addresses. But for the `spl` command it's required to comply with mainline SPL restrictions.

README.FEL.md
apritzel: I wouldn't be aware anything out there. The simplest solution is probably just a few instructions: learn the load address via `adr`, write that value right behind the code, then go to FEL. Then you can scan the SRAM via FEL to find the code and the address written. Maybe I add this to the repo some day.

paulkocialkowski:
Looks really good, thanks for writing it up!

PR #242, opened 2026-05-31
Do not try to typedef `bool` for `bare-metal.[ch]`

ijc:
Doing so results in the following with the current compiler in Debian:

```
arm-none-eabi-gcc -march=armv5te -g -Os -marm -fpic -Wall -fno-common -fno-builtin -ffreestanding -nostdinc -fno-strict-aliasing -mno-thumb-interwork -fno-stack-protector -fno-toplevel-reorder -Wstrict-prototypes -Wno-format-nonliteral -Wno-format-security jtag-loop.c bare-metal.c -nostdlib -o jtag-loop.elf -T bare-metal.lds -Wl,-N
In file included from jtag-loop.c:21:
bare-metal.h:59:13: error: 'bool' cannot be defined via 'typedef'
   59 | typedef int bool;
      |             ^~~~
bare-metal.h:59:13: note: 'bool' is a keyword with '-std=c23' onwards
```

I'd have preferred to use `stdbool.h` but that is not easily available with `-nostdinc`.

---

I don't think this is an issue with the Debian compiler, just the fact that it is new enough to have c23 support and/or this new warning.

The affected function (`uart0_puthex`) never appears to have been used from when it was introduced in 4f144b33443e7411cc7cd15436293a15f1b7d911, so I guess it could be removed, but I figured it might be useful for local debug hacking etc. Could also just remove the `newline` support I guess.

### DIFF
--- a/bare-metal.c
+++ b/bare-metal.c
@@ -594,7 +594,7 @@ void uart0_puts(const char *s)
 	}
 }
 
-void uart0_puthex(unsigned long hex, bool new_line)
+void uart0_puthex(unsigned long hex, int new_line)
 {
 	unsigned int chunks = sizeof(hex) * 2;
 	unsigned int i;
--- a/bare-metal.h
+++ b/bare-metal.h
@@ -56,7 +56,6 @@
 typedef unsigned int u32;
 typedef unsigned short int u16;
 typedef unsigned char u8;
-typedef int bool;
 
 #ifndef NULL
 #define NULL ((void*)0)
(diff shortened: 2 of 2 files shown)


ijc:
Could do #243 instead, I think I prefer this one though.

paulkocialkowski:
Please don't open two different pull requests with two different ways to solve the same issue, it is quite annoying. If you like one solution over another, just make one pull request for it and discuss other possible ways in the commit message. If maintainers disagree with it, just respin it differently.

paulkocialkowski:
That's an interesting issue. What u-boot is doing is the following:
```
NOSTDINC_FLAGS += -nostdinc -isystem $(shell $(CC) -print-file-name=include)
```

Which seems like a good way to handle it without the ugly sed part from your other commit.
Could you update this pull request in this direction and close the other one?

Thanks!

ijc:
Sorry, I fat fingered my git push and closed this PR by accidentally deleting the branch instead of updating it.

Seems like even after pushing the branch again I can't reopen. Please see #244 instead.

paulkocialkowski:
That's fine, as long as there is only one :)

PR #244, opened 2026-05-31
Use `<stdbool.h>` even on bare metal

ijc:
This is part of the compiler not the libc and so is ok to use.

This addresses this compile error with the current compiler in Debian:

```
arm-none-eabi-gcc -march=armv5te -g -Os -marm -fpic -Wall -fno-common -fno-builtin -ffreestanding -nostdinc -fno-strict-aliasing -mno-thumb-interwork -fno-stack-protector -fno-toplevel-reorder -Wstrict-prototypes -Wno-format-nonliteral -Wno-format-security jtag-loop.c bare-metal.c -nostdlib -o jtag-loop.elf -T bare-metal.lds -Wl,-N
In file included from jtag-loop.c:21:
bare-metal.h:59:13: error: 'bool' cannot be defined via 'typedef'
   59 | typedef int bool;
      |             ^~~~
bare-metal.h:59:13: note: 'bool' is a keyword with '-std=c23' onwards
```

To allow the use of `stdbool` drop use of `-nostdinc` when building bare metal
components. We will rely on the link phase and `-ffreestanding` along with
`-nostdlib` to prevent accidental use of stdlib features.

Signed-off-by: Ian Campbell <ijc@debian.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,7 @@ phoenix_info: phoenix_info.c
 	$(SUNXI_BOOT_IMAGE) $< $@
 
 ARM_ELF_FLAGS = -Os -marm -fpic -Wall
-ARM_ELF_FLAGS += -fno-common -fno-builtin -ffreestanding -nostdinc -fno-strict-aliasing
+ARM_ELF_FLAGS += -fno-common -fno-builtin -ffreestanding -fno-strict-aliasing
 ARM_ELF_FLAGS += -mno-thumb-interwork -fno-stack-protector -fno-toplevel-reorder
 ARM_ELF_FLAGS += -Wstrict-prototypes -Wno-format-nonliteral -Wno-format-security
 
--- a/bare-metal.h
+++ b/bare-metal.h
@@ -53,10 +53,11 @@
 #ifndef _BARE_METAL_H_
 #define _BARE_METAL_H_
 
+#include <stdbool.h>
+
 typedef unsigned int u32;
 typedef unsigned short int u16;
 typedef unsigned char u8;
-typedef int bool;
 
 #ifndef NULL
 #define NULL ((void*)0)


apritzel (changes requested):
Is there a problem with dropping -nostdinc, actually? I came to the conclusion that bare metal projects should keep the standard headers, to get the compiler headers, for bool, but also for stdint.h and friends. Just don't use any system *library* headers, which wouldn't work anyway unless you provide the library code as well.
Then we could just skip the fragile manual directory determination, not sure that works everywhere (MacOS? Windows? FreeBSD? LLVM?)

apritzel:
@ijc since I have you here: what about the float ABI? On Ubuntu I cannot use `arm-linux-gnueabihf` as the arm cross compiler, because that defaults to `hard`, which clashes with `-march=armv5te`, not featuring an FPU.
So I wonder if we should specify `-mfloat-abi=soft`? But I think that breaks somewhere, when there are no soft float libraries installed? Works on Ubuntu and my self-compiled cross-toolchain, though.
And should we add `-mgeneral-regs-only`, just to be safe?

ijc:
> Is there a problem with dropping -nostdinc, actually? 

I think that's for you maintainers to decide. Ultimately I just want to be able to upload an updated version of the package to Debian and have it build with the newer compiler present in current Debian.

Removing `-nostdinc` did allow me to import e..g `stdio.h` and `unistd.h` but as you say I guess actually using them would have failed to link. I don't know if there is any undesirable stuff which is available as headers only.

If coming to agreement on the way forward is going to take time can we perhaps merge this (or something like #242) while the long term plan is sorted out?

>  since I have you here

I'm afraid I have no idea about any of that sort of thing these days. The Debian packages use:
```
$(MAKE) CROSS_COMPILE=arm-none-eabi- binfiles
```
from the `gcc-arm-none-eabi` baremetal compile and that works. That compiler package is available for all the interesting hosts, including amd64, arm64 and armhf.

The sunxi-tools package appears to have been [imported unchanged into Ubuntu](https://launchpad.net/ubuntu/+source/sunxi-tools) so I assume that compile is available there too.

apritzel:
> Removing -nostdinc did allow me to import e..g stdio.h and unistd.h but as you say I guess actually using them would have failed to link

Was this just an experiment to prove that part, or is there anything behind it? I don't think we need to defend against some arbitrary code additions. And I don't think there would be ambitions to allow compiling arbitrary POSIX programs against our bare-metal "framework".

So I would lean towards staying as close to any standards as we can and need, so dropping `-nostdinc`, and avoiding anything in our bare metal code that would break (obviously).

Does anyone have easy access to a range of Linux hosts this can be tested on? I ran on Ubuntu 24.04 and Debian 13, and dropping  `-nostdinc`, but adding `-mfloat-abi=soft` worked for both the "none" and "gnueabihf" cross compiler from the standard repos. Both Debian and Ubuntu fail at the moment when the gnueabihf package was installed, as the cross-compiler detection picks that up first, and that triggers the FPU conflict with armv5te.

To be honest I consider this none cross compiler somewhat obsolete these days, since it should be a subset of the proper Linux cross-compiler. It seems to just add some C library which we cannot and don't want to use anyway.

@ijc Do you make the gcc-arm-none-eabi a build-time requirement for sunxi-tools in the Debian package?

ijc:
> Was this just an experiment to prove that part, or is there anything behind it? 

Just an experiment to see if the compiler would stop people making mistakes  in what they `#include`.

> Do you make the gcc-arm-none-eabi a build-time requirement for sunxi-tools in the Debian package?

I do, for the `binfiles` target only, which is the only one which is using `-nostdinc`.

Makefile
```diff
@@ -181,7 +181,8 @@ phoenix_info: phoenix_info.c
 	$(SUNXI_BOOT_IMAGE) $< $@
 
 ARM_ELF_FLAGS = -Os -marm -fpic -Wall
-ARM_ELF_FLAGS += -fno-common -fno-builtin -ffreestanding -nostdinc -fno-strict-aliasing
+ARM_ELF_FLAGS += -fno-common -fno-builtin -ffreestanding -fno-strict-aliasing
+ARM_ELF_FLAGS += -nostdinc -isystem $(shell $(CROSS_CC) -print-file-name=include)
```
Andre-ARM: So can we just remove that last line? Which would drop `-nostdinc`, and not bother with naming the include directory explicitly.
Which I find the more preferable solution.

Andre-ARM (changes requested):
As mentioned before, I would just drop `-nostdinc` completely.

Does that work for people?

Makefile
ijc: Done.

apritzel:
Sorry, just see that I used the wrong account above, it's the same me ;-)
Any particular reason for the patch split? To help people with bisecting? I would like to squash both...

apritzel (approved):
Thanks @ijc looks good now.
@jameshilliard can you please try this on MacOS, to see if this behaves there?
Or in general, anyone with a non-Debian setup, maybe with some non-standard cross-compiler, please test!

paulkocialkowski:
Just tried on Arch with the distro-provided compiler (`arm-none-eabi-gcc (Arch Repository) 14.2.0`), looks good!

paulkocialkowski (approved):

PR #245, opened 2026-05-31
Refresh LICENSE.md

ijc:
By running:

```console
$ curl -o LICENSE.md https://www.gnu.org/licenses/old-licenses/gpl-2.0.md
```

As well as some minor typographical changes this removes an obsolete postal address for the FSF and adjusts a non-normative example.

---

The old address triggers [a warning from the Debian package linter](https://udd.debian.org/lintian-tag/old-fsf-address-in-copyright-file), FWIW.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,14 +1,14 @@
-### GNU GENERAL PUBLIC LICENSE
+# GNU GENERAL PUBLIC LICENSE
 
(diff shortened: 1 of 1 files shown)


apritzel (approved):
Yes, indeed that's the current version from the website, and the changes are just cosmetic.

PR #246, opened 2026-05-31
Add AppStream metadata

ijc:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

This was provided by Petter Reinholdtsen in https://bugs.debian.org/1076627 and has been included in the Debian packages for a couple of years. In that bug the usefulness is described as:

> Including this information in the package will ensure programs mapping
> hardware to packages using Appstream information, like the isenkram
> package, will know that this package is useful on machines where the USB
> IDs are discovered.

---

Unsure why I didn't forward this at the time...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,9 @@ DESTDIR ?=
 PREFIX  ?= /usr/local
 BINDIR  ?= $(PREFIX)/bin
 MANDIR  ?= $(PREFIX)/share/man/man1
+APPSTREAM_METAINFODIR ?= $(PREFIX)/share/metainfo
 
-.PHONY: all clean tools target-tools install install-tools install-target-tools
+.PHONY: all clean tools target-tools install install-tools install-target-tools install-appstream-metainfo
 .PHONY: check
 
 tools: $(TOOLS) $(FEXC_LINKS)
@@ -111,6 +112,10 @@ install-misc: $(MISC_TOOLS)
 		install -m0755 $$t $(DESTDIR)$(BINDIR)/$$t ; \
 	done
 
+install-appstream-metainfo: $(APPSTREAM_METAINFO)
+	install -d $(DESTDIR)$(APPSTREAM_METAINFODIR)
+	install -m0644 org.linux_sunxi.sunxi-tools.metainfo.xml \
+		$(DESTDIR)$(APPSTREAM_METAINFODIR)
 
 clean:
 	make -C tests/ clean
--- a/org.linux_sunxi.sunxi-tools.metainfo.xml
+++ b/org.linux_sunxi.sunxi-tools.metainfo.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="console-application">
+  <id>org.linux_sunxi.sunxi-tools</id>
+  <metadata_license>MIT</metadata_license>
+  <project_license>GPL-2.0+</project_license>
+  <name>sunxi-tools</name>
+  <summary>Tools for working with Allwinner (sunxi) ARM processors</summary>
+  <description>
+    <p>This package contains various tools for working with devices
+    based around the Allwinner sunxi processors (A10/A13/A20/A31
+    etc). Utilities include tools to:</p>
+    <ul>
+      <li>interact with the processors' lowlevel bootrom (AKA FEL mode).</li>
+      <li>boot over the USB OTG port.</li>
+      <li>compile and decompile the Allwinner binary hardware
+      descriptions (FEX files).</li>
+      <li>display information about sunxi boot headers.</li>
+    </ul>
+  </description>
+  <url type="homepage">https://linux-sunxi.org</url>
+  <url type="bugtracker">https://github.com/linux-sunxi/sunxi-tools/issues</url>
+  <url type="vcs-browser">https://github.com/linux-sunxi/sunxi-tools</url>
+  <developer id="org.linux-sunxi">
+    <name>The Linux sunxi community</name>
+  </developer>
+  <provides>
+    <modalias>usb:v1F3ApEFE8d*</modalias>
+    <binary>sunxi-bootinfo</binary>
+    <binary>sunxi-fel</binary>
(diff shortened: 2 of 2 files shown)


jameshilliard:
Is this something that only makes sense to install on Linux based systems? Should it be disabled on non-Linux?

paulkocialkowski:
I don't have strong opinions about this. I honestly doubt that potential sunxi-tools users will need something like this to find out they can use it based on hardware detection. At the same time it is a FDO spec and it looks like we won't really need to update it unless we add a new binary so it probably doesn't hurt to include it.

ijc:
Disclaimer: I know next to nothing about this AppStream stuff, I'm (belatedly) forwarding a patch which was supplied to me as the Debian package maintainer.

> Is this something that only makes sense to install on Linux based systems? Should it be disabled on non-Linux?

AIUI FDO specs are generally cross platform, at least cross-Unix platform. I don't know if anything on e.g. the *BSD's exists which consumes this but I guess it's at least theoretically possible.

> I don't have strong opinions about this. I honestly doubt that potential sunxi-tools users will need something like this to find out they can use it based on hardware detection.

FWIW the patch was supplied by one of the authors of [Isenkram](https://directory.fsf.org/wiki/Isenkram) "Suggest packages to install when inserting new hardware (GUI popup)". I think possibly things like the GNOME and KDE software centers  might also support it?

But in general I share your doubt, but also:

> At the same time it is a FDO spec and it looks like we won't really need to update it unless we add a new binary so it probably doesn't hurt to include it.

Seems about right to me.

I'm not really sure that the list of the binaries is the most important bit, once there is a match on the USB device (the `<modalias>`) the whole package would generally be installed. So if the list gets a bit out of date :shrug: ...

jameshilliard:
> AIUI FDO specs are generally cross platform, at least cross-Unix platform. I don't know if anything on e.g. the *BSD's exists which consumes this but I guess it's at least theoretically possible.

I guess probably no harm adding it since it's a separate install target.

apritzel:
I don't mind having that, if it helps some people, then why not, doesn't seem to hurt. Just would be good to have someone with experience to have a look at the XML file. If no one shows up within a few days, I would just merge it anyway - since it seems to make @ijc's life easier ;-)

paulkocialkowski:
> I don't mind having that, if it helps some people, then why not, doesn't seem to hurt. Just would be good to have someone with experience to have a look at the XML file. If no one shows up within a few days, I would just merge it anyway - since it seems to make @ijc's life easier ;-)

Sounds like a plan!

Makefile
```diff
@@ -50,6 +50,9 @@ TARGET_TOOLS = sunxi-meminfo
 # Misc tools (of more "exotic" nature) not part of our default build / install
 MISC_TOOLS = phoenix_info sunxi-nand-image-builder
 
+# AppStream metadata
+APPSTREAM_METAINFO = org.linux_sunxi.sunxi-tools.metainfo.xml
```
apritzel: Do we really need this indirection? Why not just put the filename down there? I don't expect there to be multiple files in the future.

Makefile
```diff
@@ -111,6 +115,11 @@ install-misc: $(MISC_TOOLS)
 		install -m0755 $$t $(DESTDIR)$(BINDIR)/$$t ; \
 	done
 
+install-appstream-metainfo: $(APPSTREAM_METAINFO)
+	install -d $(DESTDIR)$(APPSTREAM_METAINFODIR)
+	@set -ex ; for t in $^ ; do \
```
apritzel: If we only have one file, we can skip the fancy loop.

org.linux_sunxi.sunxi-tools.metainfo.xml
```diff
@@ -0,0 +1,35 @@
...
+    <ul>
+      <li>interact with the processors' lowlevel bootrom (AKA FEL mode).</li>
+      <li>boot over the USB OTG port.</li>
+      <li>compile and decompile the Allwinner binary hardware
+      descriptions (FEX files).</li>
+      <li>display information about sunxi boot headers.</li>
+    </ul>
+  </description>
+  <url type="homepage">https://linux-sunxi.org</url>
+  <developer id="org.linux_sunxi">
```
apritzel: As linkmauve mentioned on IRC, we should use a hyphen here, to match the URL: `org.linux-sunxi`

org.linux_sunxi.sunxi-tools.metainfo.xml
```diff
@@ -0,0 +1,35 @@
...
+    etc). Utilities include tools to:</p>
+    <ul>
+      <li>interact with the processors' lowlevel bootrom (AKA FEL mode).</li>
+      <li>boot over the USB OTG port.</li>
+      <li>compile and decompile the Allwinner binary hardware
+      descriptions (FEX files).</li>
+      <li>display information about sunxi boot headers.</li>
+    </ul>
+  </description>
+  <url type="homepage">https://linux-sunxi.org</url>
```
apritzel, line 20: Since there is nothing in this file pointing to the github repo, I think we should add a line here:
`<url type="bugtracker">https://github.com/linux-sunxi/sunxi-tools/issues</url>`

apritzel (changes requested):
I have no real clue about AppStream, but found those issues when skimming over the XML file.

Makefile
ijc: I did this because it looked to be the norm and it also kept the line length more under control.

org.linux_sunxi.sunxi-tools.metainfo.xml
ijc: Done, along with adding `<url type="vcs-browser">https://github.com/linux-sunxi/sunxi-tools</url>`

apritzel: ah, great, even better, I missed that URL type

apritzel (approved):
Many thanks @ijc for the changes, looks good now.
Verified with `appstreamcli validate` to pass the test.

PR #252, opened 2026-08-01
[RFC] Add support for T153 SoC

dlan17:
Here  attempt of this PR is to add support for Allwinner T153 SoC, which tested FEL read/write/exec and bare-metal code (uart0-helloworld-sdboot.sunxi), and able to boot a mainlnie U-Boot(not public available now), the last two commits are kind of hackish which need to improve later. Mark this PR as RFC, and welcome people to further test and more comments..


soc_info.c
```diff
@@ -189,6 +189,13 @@ sram_swap_buffers a523_sram_swap_buffers[] = {
 	{ .buf1 = 0x45000, .buf2 = 0x40200, .size = 0x0400 },
 	{ .size = 0 }  /* End of the table */
 };
+
+/* Only Use SRAM_C 0x40000 - 0x67FFF */
+sram_swap_buffers t153_sram_swap_buffers[] = {
+	{ .buf1 = 0x40000, .buf2 = 0x66c00, .size = 0x1400 },
```
ssvb, line 195: Something does not add up, because `0x54000 - 0x40000 = 0x14000`. So either `.size` is missing one zero here or the commit message is inaccurate.
Is the address range 0x40000-0x40100 really reserved and breaks FEL if you overwrite it? What's the location of stacks?

dlan17: > Something does not add up, because `0x54000 - 0x40000 = 0x14000`. So either `.size` is missing one zero here or the commit message is inaccurate. Is the address range 0x40000-0x40100 really reserved and breaks FEL if you overwrite it? What's the location of stacks?

sorry, the number 0x54000 is wrong, should be 0x41400, so size = 0x1400 is correct, from my test, write to 0x40000 with 128 byte will break FEL, here is the show log

```
 $ sudo sunxi-fel version
AWUSBFEX soc=00001922(T153) 00000001 ver=0001 44 08 scratchpad=00068000 00000000 00000000
 $ stat -c %s dat2_00.bin
128
 $ sudo sunxi-fel write 0x40000 dat2_00.bin
usb_bulk_recv() ERROR -8: Overflow
```

Also from the log, spi_irq = 0x41400, sp = 0x69A00, just don't know how much space it occupy
```
 ./sunxi-fel -v spl uart0-helloworld-sdboot.sunxi
Stack pointers: sp_irq=0x00041400, sp=0x00069A00
```
And from my test, below memory space is usable with FEL protocol test
test success: 0x68000 - 0x68700
test success: 0x6B000 - 0x6C000
while I just not use the memory range of 0x68000 - 0x6C000 in this implementation, but probably we could to..

PR #255, opened 2026-09-09
Secure boot H616/A133 support

apritzel:
When an Allwinner device has the "secure boot" fuse burnt, FEL mode works differently, and we deployed some workaround code to bring FEL back into the secure world, which we rely on in the firmware.
This simple SMC workaround only worked on H3, H5, and A64, so we need new workaround code for the A133 and H616. The H6 might be in the same camp, but that needs confirmation (@smaeul?)

Factor out the existing A64 workaround code, to allow specifying multiple methods, as there will be more to come (the A523 surely needs a change at least for its GICv3).
Then the second patch introduces the code for the H616 and A133 SoCs. The assembly sequence has been developed by experimentation, mostly by comparing the CPU and system IP state between the working (non-secure) SVC mode, and the monitor mode. At the end it's really just the GIC which needs a bit or two tweaked, and switching to secure-SVC: staying in monitor mode doesn't work with the BootROM code.
The detection routine is also custom and quite different, since newer SoCs restrict less of the system when running in non-secure state. Thanks to @jameshilliard for finding the 0xa0 SID register.

I refrained from parametrising the code too much: the workaround is only applicable for a few SoCs (just two actually atm), and the GIC base address and MVBAR value plus the detection logic is the same across both. So we can hardcode this in the code, which simplifies things, and avoids fragile PC relative loads in our encoded assembly instructions.

### DIFF
--- a/fel.c
+++ b/fel.c
@@ -565,35 +565,45 @@ void aw_set_sctlr(feldev_handle *dev, soc_info_t *soc_info,
 }
 
 /*
- * Issue a "smc #0" instruction. This brings a SoC booted in "secure boot"
- * state from the default non-secure FEL into secure FEL.
- * This crashes on devices using "non-secure boot", as the BROM does not
- * provide a handler address in MVBAR. So we have a runtime check.
+ * When a SoC is configured for "secure boot", the FEL code is running in
+ * non-secure SVC mode, which denies access to secure state system registers
+ * like CNTFRQ, the secure GIC configuration registers or the RMR register to
+ * switch to AArch64.
+ * Thankfully there are ways to break out of this non-secure prison, how
+ * exactly is varying between the SoCs.
  */
-void aw_apply_smc_workaround(feldev_handle *dev)
+static void handle_secure_boot(feldev_handle *dev)
 {
 	soc_info_t *soc_info = dev->soc_info;
 	uint32_t val;
-	uint32_t arm_code[] = {
+	const uint32_t arm_smc_code[] = {
 		htole32(0xe1600070), /* smc	#0	*/
 		htole32(0xe12fff1e), /* bx	lr	*/
 	};
 
-	/* Return if the SoC does not need this workaround */
-	if (!soc_info->needs_smc_workaround_if_zero_word_at_addr)
+	/*
+	 * Check whether the SoC is using secure boot, and if the workaround
+	 * has already been applied.
+	 */
+	switch (soc_info->sec_boot_wa) {
+	case SECURE_BOOT_NONE:
 		return;
+	case SECURE_BOOT_SMC:
+		aw_fel_read(dev, soc_info->sec_mem_addr, &val, sizeof(val));
+		if (val != 0)
+			return;
+		/*
+		 * H3, H5, A64 just need a simple SMC, which will magically
+		 * return in secure state, with all the system registers
+		 * already fixed up.
+		 */
+		pr_info("Applying secure boot SMC workaround... ");
+		aw_fel_write(dev, arm_smc_code, soc_info->scratch_addr,
+			     sizeof(arm_smc_code));
+		aw_fel_execute(dev, soc_info->scratch_addr);
+		break;
+	}
 
-	/* This has less overhead than fel_readl_n() and may be good enough */
-	aw_fel_read(dev, soc_info->needs_smc_workaround_if_zero_word_at_addr,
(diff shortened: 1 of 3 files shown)


fel.c
```diff
@@ -564,36 +564,100 @@ void aw_set_sctlr(feldev_handle *dev, soc_info_t *soc_info,
...
 }
 
+static void upload_h6_secure_boot_wa(feldev_handle *dev, uint32_t scratch_addr)
+{
+	const uint32_t arm_h6_mon_code[] = {
+	htole32(0xe1600070),	/* smc	#0		; trigger SMC	*/
+	htole32(0xe3a02403),	/* mov	r2, #0x3000000	; sys resource base */
+	htole32(0xe3822a22),	/* orr	r2, r2, #0x22000; GICC base address */
+	htole32(0xe5920000),	/* ldr	r0, [r2]	; load GICC_CTLR */
+	htole32(0xe380000f),	/* orr	r0, r0, #0x6	; set ACKCTL+Grp1En */
```
jameshilliard, line 574: This encodes `orr r0, r0, #0xf`, not the commented `#0x6`, so it also sets `EnableGrp0` and `FIQEn`. Should the opcode be `0xe3800006`, or are those extra bits intentional? Please make the opcode and comment agree.

soc_info.h
```diff
@@ -99,20 +105,19 @@ typedef struct {
...
- * However it has been found out that a simple "smc" call will immediately
- * return from monitor mode, but with the NS bit cleared, so access to all
- * secure peripherals is suddenly possible.
- * The 'needs_smc_workaround_if_zero_word_at_addr' field can be used to
- * have a check for this condition (reading from restricted addresses
- * typically returns zero) and then activate the SMC workaround if needed.
+ * Depending on the SoC there are workarounds to bring us into secure state,
+ * 'secure_boot_wa' selects one of the methods. Since most workarounds must
+ * only be applied once, we need to first check if secure boot is enabled and
+ * if the workaround has already been applied. 'secure_mem_addr' holds a SoC
```
jameshilliard, line 119: These should be `sec_boot_wa` and `sec_mem_addr`, matching the actual struct fields.

soc_info.c
```diff
@@ -640,7 +645,7 @@ soc_info_t soc_info_table[] = {
 		.sid_offset   = 0x200,
 		.sid_sections = generic_2k_sid_maps,
 		.rvbar_reg    = 0x08100040,
-		.needs_smc_workaround_if_zero_word_at_addr = 0x100004,
+		.sec_boot_wa  = SECURE_BOOT_MONITOR_H6,
```
jameshilliard, line 648: The first commit's message says it drops the old workaround from H616, but the removed entry belongs to A133. Please change both H616 references in that commit message to A133.

jameshilliard (commented):
The GIC opcode and its comment disagree; there are also two documentation corrections below.

fel.c
jameshilliard: Tested against 7b713a6c8576e9a8846a1c5ce13a3ac73d5c31df on our secure-boot H616. The results support setting only AckCtl, preserving the other bits:

```c
htole32(0xe3800004), /* orr r0, r0, #4 ; set AckCtl */
```

The secure `GICC_CTLR` value immediately after SMC, before modification, was `0xb` on three cold boots. EnableGrp0, EnableGrp1 and FIQEn were already set; only AckCtl was clear.

- OR masks `0x4`, `0x6` and `0xf` each passed three cold boots, complete SID reads and five separate follow-up FEL invocations per trial. All produced `GICC_CTLR = 0xf`.
- OR masks `0x0` and `0x2`, leaving AckCtl clear, each timed out on longer FEL command sequences in all three trials.
- Deliberately clearing EnableGrp0, FIQEn, or both also passed three trials each. Clearing EnableGrp1 while retaining AckCtl failed all three longer-sequence trials. Group 1 needs to remain enabled, but the ROM already enables it.

All three successful masks also passed SPL/DRAM initialization, an unaligned 1 MiB + 3-byte write/readback, and ten subsequent FEL invocations. The AckCtl-only build additionally uploaded the full 41,082,268-byte FIT, verified all four SHA-256 hashes in U-Boot and reached `Starting kernel ...`.

A single `ver` sometimes succeeded even with the broken interrupt configuration; subsequent traffic timed out. A133 has not been hardware-tested here.

